Issue #121, opened 2011-10-15
RBENV_VERSION ignored when shelling out from ruby


bronson:
This shows that RBENV_VERSION works when executed from Perl but not from Ruby.  In this example, global version is set to 1.9.2p290.

```
$ ruby --version
ruby 1.9.2p290 (2011-07-09 revision 32553) [x86_64-linux]
$ perl -e 'system("RBENV_VERSION=system ruby -e \"puts RUBY_VERSION\"")'
1.8.7
$ ruby -e 'system("RBENV_VERSION=system ruby -e \"puts RUBY_VERSION\"")'
1.9.2
```

RBENV_VERSION should always work, shouldn't it?

mislav:
The fix of this was reverted in 95a039aaaa3855ea2df4855ad38c06faaba01f9a so I'm reopening this.

mislav:
Closing this because it's now documented as one of the pitfalls of the design of rbenv: https://github.com/rbenv/rbenv/wiki/Comparison-of-version-managers

I've had more than a decade to mull this over and there were attempts to fix, but that broke other things, so I think it's best to accept that with rbenv, once a Ruby version is activated in a specific process, the only way to switch to a different Ruby version within that process is to restore the original PATH.

Issue #14, opened 2011-08-05
Shims are incompatible with `ruby -S`


sstephenson:
```
 -S             Makes Ruby use the PATH environment variable to search for
                script, unless if its name begins with a slash.  This is
                used to emulate #! on machines that don't support it, in
                the following manner:

                      #! /usr/local/bin/ruby
                      # This line makes the next one a comment in Ruby \
                        exec /usr/local/bin/ruby -S $0 $*
```

When you pass the `-S foo` flag to Ruby, it finds `foo` in `$PATH`, then scans the first line of the file to see if it's a Ruby program and re-execs: https://github.com/ruby/ruby/blob/trunk/ruby.c#L1549-1562

The Rubinius installer uses this:

```
/Users/sam/.rbenv/versions/1.9.3-preview1/bin/ruby -S rake  -r /private/tmp/ruby-build.10166-13998/rubinius-1.2.4/config.rb -r /private/tmp/ruby-build.10166-13998/rubinius-1.2.4/rakelib/ext_helper.rb -r /private/tmp/ruby-build.10166-13998/rubinius-1.2.4/rakelib/dependency_grapher.rb build:mri
/Users/sam/.rbenv/versions/1.9.3-preview1/bin/ruby: no Ruby script found in input (LoadError)
rake aborted!
Command failed with status (1): [/Users/sam/.rbenv/versions/1.9.3-preview1/...]
```

One workaround would be to change our shims' shebang to point to another shim in our libexec directory with `ruby` in the name.  

Another workaround would be for `rbenv-exec` to prepend `$PATH` with the current version's prefix, so that children of any Ruby process will find the actual binaries in the path before our shims. The downside here is that you wouldn't be able to modify `RBENV_VERSION` in a Ruby process to spawn a child with a different version, though I'm not sure that's actually useful.

zenspider:
Looks like this should be reopened since #372 seems to have been reverted?

mislav:
No, this is not an issue presently even though that PR got reverted. The current Ruby version's bin directory gets pushed onto PATH as soon as you invoke `ruby`, so in fact the `-S` flag always has the desired effect.

The difficult problem seems to be allowing shelling out to a different version of Ruby while preserving `ruby -S` compatibility.

zenspider:
Well, I hate to break it to you, but this problem seems to have regressed:

```
cd pkg/graphics-1.0.0b4
  Successfully built RubyGem
  Name: graphics
  Version: 1.0.0b4
[...]

PR #372, opened 2013-04-01
Change `exec` semantics to allow using rbenv from a ruby process

mislav:
The most common pitfall with rbenv is that you can't shell out to another rbenv version from a ruby process started with rbenv. #121

This is an unfortunate side-effect of a PATH-modifying hack to allow support for `ruby -S`. #14

Here I propose that we stop modifying PATH during `rbenv-exec` unless we're actually dealing with `ruby -S`.

The drawback of this change is that you still can't shell out to another rbenv version from a process started with `ruby -S`. An alternative solution to this change would be to re-implement the `ruby -S` semantics and execute it without ever needing to change PATH.

### DIFF
--- a/libexec/rbenv-exec
+++ b/libexec/rbenv-exec
@@ -2,24 +2,59 @@
 #
 # Summary: Run an executable with the selected Ruby version
 #
-# Usage: rbenv exec <command> [arg1 arg2...]
+# Usage: rbenv exec <command> [args...]
 #
-# Runs an executable by first preparing PATH so that the selected Ruby
-# version's `bin' directory is at the front.
+# Runs an executable contained by the currently selected Ruby's bin
+# directory. Rough equivalent of:
 #
-# For example, if the currently selected Ruby version is 1.9.3-p327:
-#   rbenv exec bundle install
-#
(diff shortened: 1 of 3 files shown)


grosser:
can we reopen this issue plz ... ran into it again :(

evaryont:
I'm running into this problem again as well. :frowning:

grosser:
FYI this seems to work:

```Ruby
def without_rbenv(&block)
    with_env(
      "RBENV_VERSION" => nil,
      "PATH" => ENV["PATH"].split(":").reject { |x| x.include?("rbenv/versions")}.join(":"),
      &block
    )
  end

  def with_env(hash)
    old = {}
    hash.each do |k,v|
      old[k] = ENV[k]
      ENV[k] = v
    end
    yield
  ensure
    hash.each { |k, v| ENV[k] = v }
  end
```

mwpastore:
@grosser Don't you want to `ensure old.each { .. }`?

grosser:
thx, updated!

lloydwatkin:
@grosser still works in 2023 \o/

PR #377, opened 2013-04-11
added rbenv-check. checks rbenv installed and configured correctly

dyladan:
Specifically it checks that:

1. rbenv bin directory is in path
2. rbenv in bin directory is executable
3. rbenv shim directory is in path
4. `which ruby` returns the shim ruby and not system ruby

It outputs in convenient color coded format that draws attention to problems.

### DIFF
--- a/libexec/rbenv-check
+++ b/libexec/rbenv-check
@@ -1,0 +1,58 @@
+#!/usr/bin/env bash
+# Summary: Check if rbenv is installed correctly
+# Usage: rbenv check
+
+set -e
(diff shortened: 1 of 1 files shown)


mislav:
Closing in favor of [official `rbenv-doctor` script](https://github.com/rbenv/rbenv-installer/blob/40b1c0088b95fd5d9c76aa3af6ab912b7d80233b/bin/rbenv-doctor) that's being developed in a separate repo.

screen shot 2015-12-24 at 1 15 19 am

Thanks for the initial suggestions! I've initially based the logic in my script on the checks proposed above.

PR #433, opened 2013-08-07
rbenv can now initialize properly even without execution rights on $PWD

Plaristote:
With this revision, rbenv does not fail to initialize anymore when it is launched in a deleted directory or a directory that for some reason isn't accessible.

The changes just check if the old pwd is an executable file before trying to set is as the current directory.
If not, the current directory becomes the root.

Perhaps we shouldn't set any current directory if the old one is not accessible anymore.

Note that there is also warnings from rbenv-which which appear when trying to use rbenv from an unnaccessible directory. Though in such cases, rbenv still behaves normally: I didn't believe it was worth making a change.

### DIFF
--- a/libexec/rbenv
+++ b/libexec/rbenv
@@ -33,7 +33,7 @@ abs_dirname() {
   done
 
   pwd
-  cd "$cwd"
+  [ -z "$cwd" ] && cd "$cwd" || cd '/'
 }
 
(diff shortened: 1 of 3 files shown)


mislav:
Changed my mind. There wasn't evidence in the last 8 years to suggest this is worth fixing.

Thank you for the original report, though!

PR #507, opened 2013-12-18
Prevent recursive calls of `rbenv` from unnecessarily repeat-adding to PATH and RBENV_HOOK_PATH.

jf:
This also (`RBENV_HOOK_PATH`) prevents rbenv from going through the hooks for a plugin more than once for each call of `rbenv` (try running, for example, `rake -V`).


mislav:
Thank you for all your time/explanation. Closing as wontfix

PR #525, opened 2014-01-03
Avoid infinite loop in case where `pwd` returns relative path

yyuu:
The `pwd` may return relative path if the `$PWD` is badly declared in bash/zsh (e.g. `PWD="." bash`). To avoid the infinite loop in `find_local_version_file()`, stop finding the version file if the target paths are same consecutively.


Ladcjh04 (approved):

mislav:
Closing due to this being stale. I haven't heard reports of this affecting anyone else in the wild.

PR #530, opened 2014-01-13
Add installation notes for fish shell

dideler:
Also replace "Ubuntu Desktop" with "Ubuntu Bash" since the shell you use in Ubuntu matters more in this case than which version of Ubuntu.


thellimist:
https://github.com/fenekku/rbenv/blob/9a53802b34e1a3bea25db62f74dcec7725fb1f72/README.md

I think the fish part can be linked in README for easier setup for fish users. Fish users do know how to setup environment variables but it'd be a lot easier for them with documentation

PR #637, opened 2014-10-16
`rbenv version` shows symlinks acting like aliases

mislav:
Show symlinks acting like aliases under `$RBENV_ROOT/versions`:

screen shot 2014-10-16 at 12 16 41 pm

Before this symlink aliases would display just like any other version, making it hard to distinguish between "real" versions and symlinks that are just shortcuts.

:construction: Proof of concept. Needs code cleanup, tests. :construction: 

This is slated for a post-1.0 release. In discussion with @sstephenson we thought that it would be a good idea to pair this with an `rbenv alias` command to manage such aliases.
(h/t to @tpope https://github.com/tpope/rbenv-aliases)

### DIFF
--- a/libexec/rbenv-rehash
+++ b/libexec/rbenv-rehash
@@ -81,9 +81,12 @@ remove_outdated_shims() {
 
 # List basenames of executables for every Ruby version
 list_executable_names() {
-  local file
-  for file in "$RBENV_ROOT"/versions/*/bin/*; do
-    echo "${file##*/}"
+  local version file
+  rbenv-versions --bare --skip-aliases | \
+  while read version; do
+    for file in "${RBENV_ROOT}/versions/${version}/bin/"*; do
+      echo "${file##*/}"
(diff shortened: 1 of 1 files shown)


muss-web:
> Show symlinks acting like aliases under `$RBENV_ROOT/versions`:
> 
> screen shot 2014-10-16 at 12 16 41 pm
> 
> Before this symlink aliases would display just like any other version, making it hard to distinguish between "real" versions and symlinks that are just shortcuts.
> 
> :construction: Proof of concept. Needs code cleanup, tests. :construction: 
> 
> This is slated for a post-1.0 release. In discussion with @sstephenson we thought that it would be a good idea to pair this with an `rbenv alias` command to manage such aliases.
> (h/t to @tpope https://github.com/tpope/rbenv-aliases)
>

saltman007web (changes requested):
merge

PR #639, opened 2014-10-19
Support executing binstubs from user gem install location

### LINKED ISSUE #465: rbenv does not pick up executable files installed with --user-install
Archlinux has a system wide /etc/gemrc file which includes --user-install so that all gems get installed to ~/.gem/. rbenv doesn't seem to be picking up executable commands installed into ~/.gem/ruby/2.0.0/bin and does not create shims for them.

Looking at the documentation it seems that rbenv only looks for executables under ~/.rbenv/versions to generate shims. Should this not also include executables under ~/.gem ?

### LINKED ISSUE #554: Installing gems with --user-install
I'm only pointing this out because `--user-install` is the default on Arch Linux, and it took me a while to figure this one out.

When installing a gem with `--user-install`, it gets installed in `~/.gem`, where rbenv isn't aware of them or their binaries. I'm not sure rbenv can actually do anything about this, but you could point it out in the documentation.

mislav:
This brings in the functionality of [user-gems plugin](https://github.com/mislav/rbenv-user-gems).

Previously, when you did `gem install --user-install lolcat`, the "lolcat" executable will neither be found while rehashing nor available for execution. This confused users and made rbenv look like it was broken.

Caveats:
- `rbenv whence` isn't aware of user install locations
- User install locations for JRuby and Rubinius default to their default modes, i.e. this lookup isn't aware of `JRUBY_OPTS=--1.8` or `RBXOPT=-X19` or `--default-version` with which Rubinius was configured.
- Custom GEM_HOME location is not supported. If you're using a custom GEM_HOME for some reason, then you should also prepend that location's `bin/` directory to PATH. This way you skip rbenv altogether and are not dependent on its shims behavior.

Fixes #554, fixes #465

/cc @sstephenson

### DIFF
--- a/rbenv.d/which/user-gems.bash
+++ b/rbenv.d/which/user-gems.bash
@@ -5,7 +5,7 @@ if [ ! -x "$RBENV_COMMAND_PATH" ]; then
   1.9* ) gemdir="ruby/1.9.1" ;;
   2.* ) gemdir="ruby/${RBENV_VERSION:0:3}.0" ;;
   jruby-1.* ) gemdir="jruby/1.9" ;;
-  jruby-9000* ) gemdir="jruby/2.1" ;;
+  jruby-9* ) gemdir="jruby/2.1" ;;
   rbx-2.* ) gemdir="rbx/2.1" ;;
   esac
 
--- a/rbenv.d/which/user-gems.bash
+++ b/rbenv.d/which/user-gems.bash
@@ -1,0 +1,17 @@
+if [ ! -x "$RBENV_COMMAND_PATH" ]; then
+  gemdir=""
+  case "$RBENV_VERSION" in
+  1.8* ) gemdir="ruby/1.8" ;;
+  1.9* ) gemdir="ruby/1.9.1" ;;
+  2.0* ) gemdir="ruby/2.0.0" ;;
+  2.1* ) gemdir="ruby/2.1.0" ;;
+  2.2* ) gemdir="ruby/2.2.0" ;;
+  jruby-1.* ) gemdir="jruby/1.9" ;;
+  jruby-9000* ) gemdir="jruby/2.1" ;;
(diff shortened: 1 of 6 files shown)


rbenv.d/which/user-gems.bash
```diff
@@ -0,0 +1,17 @@
+if [ ! -x "$RBENV_COMMAND_PATH" ]; then
+  gemdir=""
+  case "$RBENV_VERSION" in
+  1.8* ) gemdir="ruby/1.8" ;;
+  1.9* ) gemdir="ruby/1.9.1" ;;
+  2.0* ) gemdir="ruby/2.0.0" ;;
+  2.1* ) gemdir="ruby/2.1.0" ;;
+  2.2* ) gemdir="ruby/2.2.0" ;;
+  jruby-1.* ) gemdir="jruby/1.9" ;;
+  jruby-9000* ) gemdir="jruby/2.1" ;;
```
mislav: Harcoding these paths makes me feel uneasy since it smells of not being future-proof. How do I predict which paths will RubyGems for future Ruby versions follow? I'd really love to bake this feature in but I don't want it to be half-assed or short-lived.

mislav:
I would say this is ready to :ship:, if not for one big problem: because user gem install location isn't in PATH by default, RubyGems will detect that at install time and warn the user:

```
WARNING:  You don't have /Users/mislav/.gem/ruby/2.0.0/bin in your PATH,
          gem executables will not run.
```

However this warning isn't correct with this rbenv workaround, because those executables will be discovered and run. So maybe we should ensure that this is in PATH during `gem install`? Not sure how to ensure that, other than prepending user binstub location to PATH in an `exec` hook.

mislav:
I've decided not to proceed with this for the v1.0 milestone because it's too hacky.

User gems install location will remain unsupported for now.

denisdeoniss:
1. [](url)

PR #650, opened 2014-11-08
Add option to disable warning of extraneous ruby- prefix in version

lucaspiller:
Set `RBENV_VERSION_PREFIX_NOWARN`, for example in your `~/.zshrc` or `~/.bashrc`, to disable warnings of the extraneous `ruby-` prefix in `.ruby-version` files.

rbenv's behaviour is inconsitent with other Ruby version managers. If you work on projects with others who use a Ruby version manager that prefers the `ruby-` prefix, you will see this warning every time you run a shell command in a Ruby project that has the prefix.

Further discussion of the warning exists in https://github.com/sstephenson/rbenv/issues/543, such as whether the warning should exist or not in the first place. This just adds the option to disable it, leaving the default functionality as is.

### DIFF
--- a/libexec/rbenv-version-name
+++ b/libexec/rbenv-version-name
@@ -21,9 +21,11 @@ version_exists() {
 if version_exists "$RBENV_VERSION"; then
   echo "$RBENV_VERSION"
 elif version_exists "${RBENV_VERSION#ruby-}"; then
-  { echo "warning: ignoring extraneous \`ruby-' prefix in version \`${RBENV_VERSION}'"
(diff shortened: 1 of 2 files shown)


sstephenson:
> I'm not sure if you saw the comment by RVM, but they plan on supporting other software than Ruby so would prefer to keep the prefix

…support software other than Ruby in a `.ruby-version` file? Madness.

parkr:
I prefer #687 over this – as long as the ruby version is valid, it's cool.

PR #651, opened 2014-11-09
Add shims directory to path during install

wiseley:
After following the Installation instructions, installed a ruby version and set it, `ruby -v` did not work because `shims` was not in my path. Small change here may save others some time.

Also, current version of Ubuntu Server uses .bashrc as well as Ubuntu Desktop.


parkr:
Agreed with @mislav – adding `eval $(rbenv init -)` to your shell config file should do.

PR #673, opened 2015-01-01
Added installation from source to readme

SouravMoitra:
I would like to see how to compile ruby from source and manage it with rbenv so I added a few lines

### DIFF
--- a/README.md
+++ b/README.md
@@ -267,6 +267,22 @@ hood.
 
 ### Installing Ruby Versions
 
+####From source
+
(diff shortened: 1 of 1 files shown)


mislav:
Hello, thanks for the patch, but I don't think these instructions are necessary. We don't want regular users to be compiling Ruby manually, therefore in the instructions we suggest people that they install the ruby-build plugin and build Ruby versions with `rbenv install <version>`.

If someone doesn't want to use ruby-build, but build Ruby versions manually instead, they're absolutely free to do so, but we don't want to provide guidance to that. Such person is already required to know what they're doing, or they will have to learn how to compile Ruby from Ruby's own documentation.

PR #682, opened 2015-01-20
Include gems dir in .gitignore

evaryont:
Since communal-gems is maintainer-approved, thought it would be useful to include the directory it uses in the ignore list.

(This also helps me, since I install rbenv as submodule and without this entry, the submodule is perpetually marked dirty.)


parkr:
:+1:

mislav:
Thanks!

PR #687, opened 2015-01-29
Remove warning about extraneous "ruby-" prefix in `.ruby-version`

### LINKED ISSUE #543: rbenv should not warn about "extraneous" ruby- prefix
If you use a version of `ruby-2.1.0`, rbenv emits a warning to console:

```
warning: ignoring extraneous `ruby-' prefix in version `ruby-2.1.0'
         (set by /home/eric/space/rehttp/.ruby-version)
```

As discussed [in this thread](https://github.com/rehttp/rehttp/pull/1), rbenv's behavior is not consistent with other Ruby managers like `chruby`. There should be no effect on console output if someone's trying to disambiguate their use of MRI from rubinius or jruby or something.

mislav:
People are often not able to remove the prefix in their projects because they want to support other coworkers and tools that sadly still require the prefix, like RubyMine.

So to restore sanity for a big portion of our users, the warning is now gone.

Fixes #543, closes #650

/cc @konklone @lucaspiller

### DIFF
--- a/libexec/rbenv-version-name
+++ b/libexec/rbenv-version-name
@@ -21,9 +21,6 @@ version_exists() {
 if version_exists "$RBENV_VERSION"; then
   echo "$RBENV_VERSION"
 elif version_exists "${RBENV_VERSION#ruby-}"; then
-  { echo "warning: ignoring extraneous \`ruby-' prefix in version \`${RBENV_VERSION}'"
-    echo "         (set by $(rbenv-version-origin))"
-  } >&2
   echo "${RBENV_VERSION#ruby-}"
 else
   echo "rbenv: version \`$RBENV_VERSION' is not installed" >&2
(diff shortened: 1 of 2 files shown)


parkr:
:+1:

sdhull:
:shipit:

Thanks for this, glad you all decided to remove it. Found this issue after a quick google and was happy to find the resolution. Cheers!

orangeudav:
this is not in master still ... :(

amir20:
I still get this error on the latest version `rbenv 0.4.`, is this expected? Will there be a new build?

mislav:
Please use the newest version of rbenv from master if you want this feature before the next release gets tagged.

pr0d1r2:
If anyone still has this warning (and have this file changing in your repo) you can use adaptive solution (based on https://github.com/rbenv/rbenv/issues/543#issuecomment-33978268) from https://github.com/pr0d1r2/ruby_dev_shell_aliases/blob/master/rbenv_ignore_extraneous_ruby_prefix_warning.sh which I just pushed.

PR #696, opened 2015-02-16
Update README to make rbenv version consistent with previous `versions` output

aprescott:
The `rbenv versions` info highlights 1.9.3-p327 as the current version, and making `rbenv version` sample output match that helps make it clearer.


mislav:
I'm usually against constantly bumping version numbers in examples to latest, but in this case it's making the example consistent with other ones in the same document so :+1: good catch

aprescott:
Thanks!

PR #700, opened 2015-03-02
minor: also remove (added) leading ":" in `remove_from_path`

blueyed:


libexec/rbenv-which
```diff
@@ -23,7 +23,7 @@ remove_from_path() {
     path_before="$result"
     result="${result//:$path_to_remove:/:}"
   done
-  echo "${result%:}"
+  echo "${result%:#:}"
```
mislav: You can't combine string substitutions like that. On my stock bash 3.2.57, this does no substitution whatsoever.

mislav:
Doesn't work as expected, and I don't think it's really worth fixing all that much.

PR #707, opened 2015-03-16
README: rbenv is a function -> rbenv is a shell function

adelevie:
```
➜  .rbenv git:(master) type rbenv
rbenv is a shell function
```


mislav:
See https://github.com/sstephenson/rbenv/pull/762#issuecomment-126520056

Issue #369, opened 2013-03-28
The master branch of rbenv breaks inside of tmux.


tmiller:
When using the master branch version of rbenv, tmux no longer respects rbenv. For example I have global ruby setup to 2.0.0-p0 and when I run `rbenv version` it correctly reports 2.0.0-p0, it is also correctly reported when I run ruby --version. Now in the same session I start a new tmux session using `tmux new`. Now when I run `rbenv version` it still correctly reports 2.0.0-p0, however when I run `ruby --version` it now reports 1.8.7. When I run `which ruby` it shows me `/usr/bin/ruby` instead of the ruby shim.

PR #722, opened 2015-04-19
Fix test for adding shims in fish

blueyed:
Commit e2173df4 (for issue #369) did not handle the fish test properly.

This renames it and fixes the assertion.


mislav:
Thanks! I guess my assertion was botched so it passed somehow.

blueyed:
> I guess my assertion was botched so it passed somehow.

Yes.

PR #725, opened 2015-04-29
rbenv-init: add "toward the end" to install instructions

blueyed:
This was suggested in https://github.com/yyuu/pyenv/pull/313.


mislav:
How about changing "adding" to "appending", which automatically means towards the end?

blueyed:
Makes sense.
Will you do this change yourself, or should I amend the PR?

PR #730, opened 2015-05-10
Fix shell version when invoked from a script

blueyed:
When invoked from a shell script, `$(rbenv init -)` did not get the shell name correct.
It needs to look at the `args` value from `ps`.

Reported for pyenv at: https://github.com/yyuu/pyenv/issues/373

### DIFF
--- a/libexec/rbenv-init
+++ b/libexec/rbenv-init
@@ -22,7 +22,7 @@ done
 
 shell="$1"
 if [ -z "$shell" ]; then
-  shell="$(ps c -p "$PPID" -o 'ucomm=' 2>/dev/null || true)"
+  shell="$(ps -p "$PPID" -o 'args=' 2>/dev/null || true)"
   shell="${shell##-}"
   shell="${shell%% *}"
   shell="$(basename "${shell:-$SHELL}")"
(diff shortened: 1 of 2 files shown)


mislav:
Hi, I don't get what use case does this solve. `rbenv init` shell detection was meant for then you eval its output in an interactive shell, and it detects what type is that shell. Why would you run `rbenv init` from a script?

blueyed:
> Why would you run rbenv init from a script?

I've used this on a server IIRC.

mislav:
On the server, presumably in a non-interactive shell, the preferred approach is to prepend `` `rbenv root`/shims`` to PATH manually instead of invoking `rbenv init` which is only meant for interactive shells.

Maybe I should detect the interactive state of the current shell and simply not execute any commands that are meant for interactive shells, such as loading completions or defining the `rbenv()` function?

mislav:
I'm thinking of closing this as wontfix. I can't reproduce your problem. Your added tests passes even without your patch.

mislav:
Ah, I see: your test passes without patch on OS X (my laptop), but fails on Linux https://travis-ci.org/rbenv/rbenv/jobs/98663235.

There are many differences between BSD `ps` and POSIX `ps`, unfortunately.

PR #735, opened 2015-05-17
rbenv-rehash: look for program==ruby*

blueyed:
When "program" is not ruby, but e.g. ruby2.1, it should be handled
specially, too.

I came across this for pyenv.
Some discussion: https://github.com/yyuu/pyenv/pull/368#issuecomment-102810231

### DIFF
--- a/libexec/rbenv-rehash
+++ b/libexec/rbenv-rehash
@@ -46,7 +46,7 @@ set -e
 [ -n "\$RBENV_DEBUG" ] && set -x
 
 program="\${0##*/}"
-if [ "\$program" = "ruby" ]; then
+if [[ "\$program" =~ ^ruby([0-9][.0-9]+)?$ ]]; then
   for arg; do
     case "\$arg" in
     -e* | -- ) break ;;


mislav:
Thanks for bringing it to my attention, but not sure about whether I want to support this. Nobody has come up against this in a real-world scenario. Why would someone suffix their `ruby` executable with a version number if they're not installing under a shared location like `/usr/bin`, but under `RBENV_ROOT/versions`?

blueyed:
I guess that's more likely to be the case (or rather only relevant) with pyenv, where you can have multiple versions being active at the same time, e.g. "python3" and "python2".

mislav:
JRuby historically could execute in different compatibility modes (Ruby 1.8 vs. 1.9 vs. 2.0) but that was decided using command-line flags or environment variables.

I'm going to close this because the only time people choose to suffix the ruby executable is when they are installing in a shared location. This isn't the same case as when rbenv is managing versions.

We can reopen if someone reports that they've hit against this.

blueyed:
Please consider re-opening it: there is more information with the issue for pyenv (where it was merged): https://github.com/yyuu/pyenv/pull/379#issuecomment-120716518

mislav:
What's the exact issue with symlinks and how does it relate to comparing the program name to "ruby" or "ruby*"?

PR #739, opened 2015-05-28
Expose version-name and version-origin hooks

jasonkarns:
- version-name hook
  It is invoked *after* the traditional `RBENV_VERSION` lookup. Which means hook scripts can interrogate `$RBENV_VERSION_FILE` and/or `$RBENV_VERSION` (or use the executables).

  The hooks are then run, giving plugins a chance to alter `RBENV_VERSION`. Once the hooks have run, we now have (in `$RBENV_VERSION`) the actual version we want to use (or it's empty which defaults to `system` per normal). Lastly, the same logic remains for checking if the version exists, or trimming the `ruby-` prefix.
- version-origin hook

  It is invoked *before* the traditional `rbenv-version-file` lookup. Because `version-origin` is usually run immediately after `version-name`, then any plugin hooks that alter `version-name` would have done so. Thus, running `version-origin` prior to printing the origin gives those plugins a chance to alter the `version-origin` to match. If any of the hooks set `$RBENV_VERSION_ORIGIN`, then it is used as the return value. Otherwise, the existing logic continues to return "environment variable" or "filename" as appropriate.

With both of these hooks in place, we have a clean seam by which plugins can inject their own ruby-version-setting logic. Until now, the only mechanism to available to alter the chosen ruby version was in the `which` hook. Using the `which` hook is error prone, however, since the hooks must build up the full
`RBENV_COMMAND_PATH` themselves. Further, with the `which` hook, commands like `rbenv version` will report a different ruby than they one that will actually be executed. And functions like `__rbenv_ps1` report the wrong ruby version.

Using these new hooks, plugins can safely manipulate the version of ruby chosen, while still leaving the executable-path-lookup to rbenv.

### DIFF
--- a/libexec/rbenv-hooks
+++ b/libexec/rbenv-hooks
@@ -9,6 +9,8 @@ set -e
 if [ "$1" = "--complete" ]; then
   echo exec
   echo rehash
+  echo version-name
+  echo version-origin
   echo which
   exit
 fi
--- a/libexec/rbenv-version-name
+++ b/libexec/rbenv-version-name
@@ -8,6 +8,13 @@ if [ -z "$RBENV_VERSION" ]; then
   RBENV_VERSION="$(rbenv-version-file-read "$RBENV_VERSION_FILE" || true)"
 fi
 
+OLDIFS="$IFS"
+IFS=$'\n' scripts=(`rbenv-hooks version-name`)
+IFS="$OLDIFS"
+for script in "${scripts[@]}"; do
+  source "$script"
+done
+
 if [ -z "$RBENV_VERSION" ] || [ "$RBENV_VERSION" = "system" ]; then
   echo "system"
   exit
--- a/libexec/rbenv-version-origin
+++ b/libexec/rbenv-version-origin
@@ -3,7 +3,16 @@
 set -e
 [ -n "$RBENV_DEBUG" ] && set -x
 
-if [ -n "$RBENV_VERSION" ]; then
+OLDIFS="$IFS"
+IFS=$'\n' scripts=(`rbenv-hooks version-origin`)
+IFS="$OLDIFS"
+for script in "${scripts[@]}"; do
+  source "$script"
+done
+
+if [ -n "$RBENV_VERSION_ORIGIN" ]; then
+  echo "$RBENV_VERSION_ORIGIN"
+elif [ -n "$RBENV_VERSION" ]; then
   echo "RBENV_VERSION environment variable"
 else
   rbenv-version-file
--- a/test/hooks.bats
+++ b/test/hooks.bats
@@ -2,11 +2,6 @@
 
 load test_helper
 
-create_hook() {
-  mkdir -p "$1/$2"
-  touch "$1/$2/$3"
-}
-
 @test "prints usage help given no argument" {
   run rbenv-hooks
   assert_failure "Usage: rbenv hooks <command>"
@@ -15,11 +10,13 @@ create_hook() {
 @test "prints list of hooks" {
   path1="${RBENV_TEST_DIR}/rbenv.d"
   path2="${RBENV_TEST_DIR}/etc/rbenv_hooks"
-  create_hook "$path1" exec "hello.bash"
-  create_hook "$path1" exec "ahoy.bash"
-  create_hook "$path1" exec "invalid.sh"
-  create_hook "$path1" which "boom.bash"
-  create_hook "$path2" exec "bueno.bash"
+  RBENV_HOOK_PATH="$path1"
+  create_hook exec "hello.bash"
+  create_hook exec "ahoy.bash"
+  create_hook exec "invalid.sh"
+  create_hook which "boom.bash"
+  RBENV_HOOK_PATH="$path2"
+  create_hook exec "bueno.bash"
 
   RBENV_HOOK_PATH="$path1:$path2" run rbenv-hooks exec
   assert_success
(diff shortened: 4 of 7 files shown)


jasonkarns:
As an example use-case of these new hooks, here are proposed changes to the [rbenv-bundler-ruby-version](https://github.com/aripollak/rbenv-bundler-ruby-version) plugin: https://github.com/aripollak/rbenv-bundler-ruby-version/compare/master...jasonkarns:version-hooks

The changes allow the plugin hook into `version-name` hook and pull the ruby version from a Gemfile. (If not set already via `rbenv shell` or `rbenv local`) Additionally, the plugin hooks into `version-origin` to report that the ruby version was set via "Gemfile". A nice side effect of these new hooks (and the updates to the plugin they enable), is that this plugin will no longer break the [rbenv-binstubs](https://github.com/ianheggie/rbenv-binstubs) plugin. (They are currently incompatible with one another because they both attempt to override `$RBENV_COMMAND_PATH` via the `which` hook.)

mislav:
I dig the functionality. The only problems I see with this are:

1. The potential slowdown. We're adding two more hook lookups to every rbenv or ruby invocation. Searching the RBENV_HOOK_PATH for hook scripts is not cheap, even if none are found (clean install). I will measure the speed difference with this PR and report back.

2. The fact that it's two separate hooks. So if a Gemfile plugin wanted to support the `ruby "..."` declaration in a Gemfile, it would need to publish two separate hooks. But, it's likely that the two scripts would be performing largely similar work. In your example diff above, the two hooks both call `rbenv-bundler-ruby-version`— once for `version-name`, another time to display "Gemfile" for `version-origin`. The two invocations will contribute to rbenv being slower.

Is it possible that we could merge this functionality in a single hook from which you would be able to export two different variables: `RBENV_VERSION` and `RBENV_VERSION_ORIGIN`?

jasonkarns:
Yeah, I'm concerned about the performance implications and was hoping that you might already have some facilities for measuring the changes :)

> We're adding two more hook lookups to every rbenv or ruby invocation.

This actually only adds 2 hook lookups to the `rbenv version` command. It's the only command that fires both `rbenv-version-name` *and* `rbenv-version-origin`. All other commands (like `exec`) invoke only `rbenv-version-name`, so only the `version-name` hook would be fired.

But in principle, I agree, a single hook would be better. Since the `name` hook fires *after* most of the existing version lookup, and the `origin` hook fires *before* the existing origin lookup, they could both be merged. Some potential gotchas if they were to be merged:

1. The `version-name` script is currently invoked in 4 additional locations separately from `version-origin` (`exec`, `prefix`, `versions`, `which`). So *if* the additional work to compute the origin is expensive, then it would be spent in at least these 4 commands where the result is never even used. Of course this additional cost would only be born by those who choose to use expensive plugins.

2. `version-origin` is invoked alone (without the corresponding `version-name`) in `rbenv-local`. Which means this would be the only location where the origin is looked up without the hook having run. I don't believe there is any issue here but I'm not sure. This could lead to inconsistencies.

So while I agree that invoking (likely) the same type of code twice is theoretically bad, the two hooks are only invoked together from the `rbenv version` command. Which is only invoked manually–never by other commands or as part of normal ruby command usage. Further, the plugins' hook scripts could probably memoize the result so it's only done once. (haven't proven this)

I think the fact that `version-name` and `version-origin` are already separate plumbing commands lends credence to the idea that the hooks should remain separate. They are accomplishing different tasks for different use-cases, even though they (may) incidentally require the same type of computation.

mislav:
This PR adds 4~5ms to the overhead of execution of rbenv shims. That's the overhead which will apply to every time you call `ruby`, `bundler`, etc.

```
$ measure 100
mean:    0.044
stdev:   0.005

$ git checkout master
Switched to branch 'master'
Your branch is up-to-date with 'origin/master'.

$ measure 100
mean:    0.040
stdev:   0.004
```

So on my MBP with bash C extension compiled (a feature of rbenv in master), the overhead increases from about 40ms to 45ms. Not terrible, but lets keep in mind that this applies to everything you do with rbenv. And with an actual `version-name` plugin, especially if it was badly written, people's experience with rbenv might degrade again.

On the other hand, I like the feature and really want to keep it. I thought some more about consolidating the `version-name` and `version-origin` hooks into one hook, but it would require restructuring of rbenv internals right now. And, `version-origin` doesn't really get called unless you ask for `rbenv version`, which I'm fine if it's not terribly fast. The most important thing to keep fast is the shim execution, the primary function of rbenv, which only calls out to `rbenv-version-name` right now.

jasonkarns:
Two questions on your tests:

1. Did you have an actual version-name hook supplied? Or is the additional 5ms exclusively from searching for hooks to execute?

2. I imagine that the speed of the hook search is proportional to the number of plugin directories it must search? How many plugin directories were searched for these tests? And do you have any idea how much of the hook-script search is constant time vs how much is a marginal increase per additional plugin?

> but it would require restructuring of rbenv internals right now

Agreed. Keeping the name and origin lookups separate would keep the shim execution as fast as possible, both with and without the hooks.

mislav:
On 29 May 2015, at 17:12, Jason Karns <notifications@github.com> wrote:
> Did you have an actual version-name hook supplied? Or is the additional 5ms exclusively from searching for hooks to execute?
> 
No hook; just lookup
> I imagine that the speed of the hook search is proportional to the number of plugin directories it must search? How many plugin directories were searched for these tests?
> 
Default directories of RBENV_HOOK_PATH: ~5 to start with, then increases for each plugin installed.

hurrymaplelad:
+1 This would help me write plugins.

PR #742, opened 2015-06-05
Don't change global git configuration. fixes #741

### LINKED ISSUE #741: Running tests changes global git user name/email
After running the tests (`bats test/`), my global git user name and email is wrong. `test/--version.bats` sets the name and email to Tester <tester@test.local>.

https://github.com/sstephenson/rbenv/blob/5b9e4f05846f6bd03b09b8572142f53fd7a46e62/test/--version.bats#L7-L8

Am I running the tests incorrectly or is there some configuration that I should be doing to prevent this from happening? (Temporarily, I've just removed those lines from the file and have marked them as `--assume-unchanged` in the git index)

jasonkarns:
`test/--version.bats` needs to create a local fake clone of rbenv in order to test the `--version` flag. In order to set up the test scenario correctly, the fake rbenv clone needs to have some commits and tags, which means a user (name/email) must be set.

Previously, this test was relying on the fact that the test suite operates under a different `$HOME`. This meant that the global git configuration was being set only in `$RBENV_TEST_DIR/home/.gitconfig`. However, git has supported the XDG BaseDir spec for some time now. Which means that git will prefer `$XDG_CONFIG_HOME/git/config` over `$HOME/.gitconfig`. Which meant that running this test suite on a system with XDG_CONFIG_HOME set would alter the user's actual user name and email.

One solution is to change the `$XDG_CONFIG_HOME` environment variable during the test run. However, in order to fully insulate the test suite from XDG setups, we would *also* have to overwrite the `$XDG_CONFIG_DIRS` var. This is an aweful lot of environment tweaking when all we really want to do is get some commits into our fake rbenv git repo.

So, rather than continue to monkey with the global git config, this PR simply sets the user's name/email *local to* the git repo. (That is, under `$RBENV_TEST_DIR/root/.git/config`, which is done by default for any git-config command without the `--global` flag.) This way the temporary git configuration is kept local to the temporary git repo itself.

In fact, the user name/email may not even be necessary at all. Current versions of git emit a warning message when a git-commit is made while the user isn't set. However, the commit itself still succeeds and returns a successful status code. This PR doesn't rely on this behavior, however, because it's possible that older versions of git treat an un-set user as an actual error. But this remains a potential cleanup–depending on which versions of git we support (and if any version of git actually treats an un-set user as an actual error.)

In addition to making the git config setting local (which meant moving the commands from setup to after the git-init call), I factored out a function that does all of the git setup (creating the repo directory, changing dirs into it, initializing the git repo, and configuring the user).

If it weren't for the `default` test, all of these steps could have remained in `setup`.

#741

### DIFF
--- a/test/--version.bats
+++ b/test/--version.bats
@@ -2,10 +2,9 @@
 
 load test_helper
 
-setup() {
-  mkdir -p "$HOME"
-  git config --global user.name  "Tester"
-  git config --global user.email "tester@test.local"
+git_config_user() {
+  git config user.name  "Tester"
+  git config user.email "tester@test.local"
 }
 
 git_commit() {
@@ -23,6 +22,7 @@ git_commit() {
   mkdir -p "$RBENV_ROOT"
   cd "$RBENV_ROOT"
   git init
+  git_config_user
   git_commit
   git tag v0.4.1
   git_commit
@@ -38,6 +38,7 @@ git_commit() {
   mkdir -p "$RBENV_ROOT"
   cd "$RBENV_ROOT"
   git init
+  git_config_user
   git_commit
 
   cd "$RBENV_TEST_DIR"
--- a/test/--version.bats
+++ b/test/--version.bats
@@ -2,7 +2,10 @@
 
 load test_helper
 
-git_config_user() {
+setup_rbenv_git_clone(){
+  mkdir -p "$RBENV_ROOT"
+  cd "$RBENV_ROOT"
+  git init
   git config user.name  "Tester"
   git config user.email "tester@test.local"
 }
@@ -19,10 +22,8 @@ git_commit() {
 }
 
 @test "reads version from git repo" {
-  mkdir -p "$RBENV_ROOT"
-  cd "$RBENV_ROOT"
-  git init
-  git_config_user
+  setup_rbenv_git_clone
+
   git_commit
   git tag v0.4.1
   git_commit
@@ -35,10 +36,8 @@ git_commit() {
 }
 
 @test "prints default version if no tags in git repo" {
-  mkdir -p "$RBENV_ROOT"
-  cd "$RBENV_ROOT"
-  git init
-  git_config_user
+  setup_rbenv_git_clone
+
   git_commit
 
   cd "$RBENV_TEST_DIR"


mislav:
Thanks for the PR. How about `unset XDG_CONFIG_HOME` in `test_helper`, and then not changing any of the other test code?

jasonkarns:
Would also have to unset `$XDG_CONFIG_DIRS` and potentially other xdg related variables. Which just becomes a game of whack a mole attempting to guard against global side effects. Attempting to mess with global variables simply to allow setting a global config option is a really brittle approach (as evidenced by #741). And that's fine, if there weren't an easier way. Setting the git config local to the git repo to which it applies is a cleaner (IMO) solution with zero potential impact on global state/config and no reliance on global environment variables.

jasonkarns:
I think an even better solution is completely eliminating the `git config user` invocations entirely. The git commits are to a temporary git repo that ends up being wiped anyhow. So the only reason the user even needs to be set is to avoid warning messages. And the warning messages can be ignored anyway since `git commit` still returns a 0 exit status with an un-set user. This all seems to be a lot of contortions and environment changing simply to avoid a warning message that doesn't impact the test.

Adding `--quiet` to the git commit function would silence the warning message.

mislav:
The test suite generally tries to be isolated from the user's environment. Resetting as many environment variables as possible is a good thing because it brings us closer to that.

I encourage you to try this approach:

```
for xdg_var in `env | grep ^XDG_ | cut -d= -f1`; do
  unset "$xdg_var"
done
```

jasonkarns:
Instead of isolation-through-nuking, wouldn't it be safer and more future
proof to be isolated by not doing things that are impacted by the
environment?

mislav:
I don't think so, because I don't know which other commands are impacted by the environment. Case in point: I didn't know that `git config` is affected by `XDG_*` configuration. Now, to remedy this, you're telling me "hey let's just stop using `git config`". But my approach would be to reset these offending variables. Because, if they affect git config, then they might affect other commands in tests (that either exist right now in the suite or might get added in the future).

mislav:
I guess it all comes down to: I *want* to be able to use `git config --global` with confidence that I'm going to be writing in an isolated, test-only temporary HOME directory. I want to nuke all environment that leaked that might stand in my way. In the future, I might need `git config --global` for things other than appeasing `git commit`.

jasonkarns:
> I want to nuke all environment that leaked that might stand in my way. 

Taking this to the extreme... In order to be sure that there aren't other unknown environment variables in the future, the extreme step would be to nuke everything in `env`

> In the future, I might need git config --global for things other than appeasing git commit.

Eh, I think the test relying on something in the global is the code smell in the first place. There's nothing special about the global config that couldn't be set locally. (I know this was a hypothetical, though)

If the test directory is meant to be an isolated test environment, then restricting the configuration to its local settings is pretty well isolated. An alternative to this whole thing is creating a properly configured git repo as a test fixture and keeping it versioned as a regular test fixture would be.

So options:

1. on-demand config setup, locally (this PR)
2. no env resetting or config settings at all, and simply mute the warnings with `--quiet`
3. git repo as test fixture (no on demand configuration in the test)
4. nuking global environment settings

I'm mostly a fan of 2. Barring that, 1. If you want to do 4, that's your prerogative. But I'm not putting my hands on that :)

PR #756, opened 2015-07-17
Display version origin with non-installed versions

blueyed:
This is useful as an indicator where it is coming from.


mislav:
Thanks! I agree this might help people understand where a local version has been set from.

PR #758, opened 2015-07-24
better format

wankai:


mislav:
Thanks, but this was an experimental feature, and I never added the indent in case we will drop the surrounding conditional and keep the old code. We might format it better in the future, but for now, I think the way it is now is alright.

PR #760, opened 2015-07-25
Add configuration for zsh user

SamuraiT:
If zsh users add only this:

~~~sh
$ echo 'eval "$(rbenv init -)"' >> ~/.zshrc
~~~
you will get this error

~~~sh
/usr/local/Cellar/rbenv/0.4.0/libexec/../completions/rbenv.bash: command not found: complete
~~~
by adding `zsh` after `-`, we can solve this problem :)

~~~sh
$ echo 'eval "$(rbenv init - zsh)"' >> ~/.zshrc
~~~

### DIFF
--- a/README.md
+++ b/README.md
@@ -178,7 +178,13 @@ easy to fork and contribute any changes back upstream.
     $ echo 'eval "$(rbenv init -)"' >> ~/.bash_profile
     ~~~
 
-    _Same as in previous step, use `~/.bashrc` on Ubuntu, or `~/.zshrc` for Zsh._
+    _Same as in previous step, use `~/.bashrc` on Ubuntu_
+    
+    For zsh users, add `rbenv init` to your zsh.
+    
+    ~~~ sh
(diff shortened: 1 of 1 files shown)


mislav:
Are you using latest version of rbenv from master? This parameter shouldn't be needed anymore.

SamuraiT:
oh really ? I just brew installed rbenv. Then I will update, Thanks

SamuraiT:
hmm I updated but still it shows me the error. So I think people who installed via brew and zsh users need to add `- zsh`.

thomasjo:
@SamuraiT You **really** should install rbenv by cloning the repository. The last release (v0.4.0) was almost 2.5 years ago. AFAIK the recommended way to install rbenv is via git.

SamuraiT:
oh ok I didn't know it. Thank you I will install from git then.

SamuraiT:
Since installing rbenv via git is recommended, I will close this PR

PR #762, opened 2015-07-30
Refresh the returned wording after installation

cyril:


mislav:
Bash returns "rbenv is a function". Even though your proposed wording is more descriptive, I think we should match what bash outputs because it's the primary shell that most people will have.

PR #806, opened 2015-10-25
Avoid running `rbenv rehash` multiple times during `bundle install`

mislav:
This is an attempt to work around the fact that Rubygems post_install hooks may happen multiple times per single `bundle install` and ideally we want `rbenv rehash` to run only once after all gems have been installed.

### DIFF
--- a/rbenv.d/exec/gem-rehash/rubygems_plugin.rb
+++ b/rbenv.d/exec/gem-rehash/rubygems_plugin.rb
@@ -15,13 +15,19 @@ if defined?(Bundler::Installer) && Bundler::Installer.respond_to?(:install)
     class << self
       alias install_without_rbenv_rehash install
       def install(root, definition, options = {})
-        result = install_without_rbenv_rehash(root, definition, options)
         begin
-          if result && Gem.default_path.include?(Bundler.bundle_path.to_s)
-            `rbenv rehash`
+          if Gem.default_path.include?(Bundler.bundle_path.to_s)
+            bin_dir = Gem.bindir(Bundler.bundle_path.to_s)
+            bins_before = File.exist?(bin_dir) ? Dir.entries(bin_dir).size : 2
           end
         rescue
           warn "rbenv: error in Bundler post-install hook (#{$!.class.name}: #{$!.message})"
+        end
+
+        result = install_without_rbenv_rehash(root, definition, options)
+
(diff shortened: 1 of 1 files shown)


rbenv.d/exec/gem-rehash/rubygems_plugin.rb
```diff
@@ -6,13 +6,38 @@
...
+        begin
+          if Gem.default_path.include?(Bundler.bundle_path.to_s)
+            bin_dir = Gem.bindir(Bundler.bundle_path.to_s)
+            bins_before = File.exist?(bin_dir) ? Dir.entries(bin_dir).size : 2
+          end
+        rescue
+          warn "rbenv: error in Bundler post-install hook (#{$!.class.name}: #{$!.message})"
+        end
+
+        result = install_without_rbenv_rehash(root, definition, options)
```
sandlerr: @mislav this appears to cause an endless loop due to the recursive method call (the alias just calls install over and over)

sandlerr: This snippet demonstrates the unexpected behaviour of `alias` in these circumstances

```
class Test
  class << self
    def output
      puts 'original'
    end
  end
end


Test.class_eval do
  class << self
    alias original output
    def output
      original
      puts 'not original!'
    end
  end
end


Test.output
```

mislav: Thanks for reporting. Does the endless loop happen to you when you run `bundle install`? If not, which command did you run? What is your `gem --version`? What is your `ruby --version`? Thanks!

MaxPower15: @mislav I also ran into this and it was kind of tough to reproduce from the terminal. I have a ruby daemon that runs a chef deploy that uses the rbenv chef cookbook to shell out to rbenv. So it's very hard to reproduce the exact environment. I was able to fix by adding `&& !Bundler::Installer.respond_to?(:install_without_rbenv_rehash)` to line 13.

To answer your questions though:

```
$ ruby -v
ruby 2.1.2p95 (2014-05-08 revision 45877) [x86_64-linux]
$ gem -v
2.2.2
$ bundle -v
Bundler version 1.10.6
```

mislav: @MaxPower15 Yes, please submit a PR!

PR #809, opened 2015-10-26
Make sure the alias statement only executes once

sandlerr:
This is the fix suggested by @MaxPower15 [here](https://github.com/sstephenson/rbenv/pull/806/files#r43037156) to fix the infinite loop edge case of #806.


MaxPower15:
Woo, beat me to it! Thanks. :) I've been trying to figure out how to make a test for this but didn't see an obvious place to put it. It seems to me like this is a pretty safe change though, right? I can't imagine why we would want to allow it to be redefined several times.

Issue #311, opened 2013-01-06
Failed to create hard link `erb'


kud:
Hello,

Well, first of all, I'd like to explain the context because I'm in a "special" environment.

I'm on Windows and I use an Ubuntu server as virtual machine (with virtual box). I've got a shared folder between my host and my guest which is my user home directory.

Maybe you know but for safety reasons, you can't make any symlink on your shared folder, except if you enable the option in virtual box, thing I did. It now perfectly works with npm for instance which requires some symlinks.

I use Kitty to connect to my VM via SSH and I'm able to do whatever I want on my windows shared folder via a linux environment, which is great.

Though, It seems that I can't install ruby via rbenv / ruby-build: I've got this error:

```
ln: failed to create hard link `erb' => `/home/host/.rbenv/shims/.rbenv-shim': Operation not permitted
```

Apparently, make hard links are a bad idea: 
- http://www.cyberciti.biz/faq/creating-hard-links-with-ln-command/
- http://unix.stackexchange.com/questions/22394/why-hard-links-not-allowed-to-directories-in-unix-linux

Why don't you do soft link (ln -s)? like what homebrew or npm do?

Do you have any idea how to fix this bug?

Thank you a lot!

PR #814, opened 2015-10-28
Never use hardlinks. Never.

### LINKED ISSUE #64: Rehash fails on network device where hard links are unsupported
FHiMac2:.rbenv alexa$ rbenv rehash
    ln: erb: Operation not supported

This is after following the instructions as written in the readme and using ruby-build.
Please let me know what other information you may need.

m-o-e:
Make shim copies instead of hardlinks.

Copies consume more diskspace than hardlinks,
but we are talking *kilobytes*. Nobody will notice the difference.

Hardlinks cause serious problems.

They fail:
- On network fileystems (NFS)
- On windows
- On docker (aufs and overlay mounts)

These are common deployment scenarios.
Increasingly so with docker and virtualized runtimes.

rbenv is supposed to make life easier, not harder.
Please remove this misguided optimization.

Fixes #64 #311 #312

### DIFF
--- a/libexec/rbenv-rehash
+++ b/libexec/rbenv-rehash
@@ -113,7 +113,7 @@ install_registered_shims() {
   local shim file
   for shim in $registered_shims; do
     file="${SHIM_PATH}/${shim}"
-    [ -e "$file" ] || ln -f "$PROTOTYPE_SHIM_PATH" "$file"
+    [ -e "$file" ] || cp "$PROTOTYPE_SHIM_PATH" "$file"
   done
 }
 


mislav:
Thanks for the suggestion. I agree that it would be simpler to avoid hard links. However, I think that hard links might have been used to speed up rehash process, not save disk space. @sstephenson can confirm.

Here's `rbenv rehash` performance on the master branch:

```
mean:    0.347
stdev:   0.003
```

Here's with your proposed change:

```
mean:    0.367
stdev:   0.003
```

There's a ~20ms slowdown. This is in an environment that has 117 shims and lots of Ruby versions. For average users, this slowdown would probably be much less.

I might be down with taking a performance hit for greater portability. With C extension compiled to speed up `realpath` calls, the final measurement including your change is:

```
$ make -C src
...

mean:    0.306
stdev:   0.006
```

Not too bad.

* * *

One other thing we could also try is doing `ln- f`, but fallback to `cp` if that fails. Not sure if this could have negative repercussions, though.

m-o-e:
Thanks for your consideration.

Yes, a small performance hit is to be expected on most filesystems.
So small as to be negligible, unless your rbenv is backed by a [1530 C2N](https://en.wikipedia.org/wiki/Commodore_Datasette).

> One other thing we could also try is doing ln- f, but fallback to cp if that fails

Not worth it, IMHO.

You're still buying kilobytes and wallclock-milliseconds
at the cost of uncertain future failure modes, *which
have already bitten you once*.

It would merely add insanity (cf. Einstein)
to your premature optimization (cf. Knuth).

sstephenson:
+1. Hard-linking was for performance, not space savings, but a 5% speed improvement is not worth the tradeoffs. Although I do remember the difference being much more pronounced four years ago.

mislav:
@sstephenson Thanks for your input!

PR #818, opened 2015-11-06
Don't export `$PATH`

geoffharcourt:
`PATH` does not need to be exported to be available to the shell. This
change removes instances where `PATH` is exported in `rbenv` and
`rbenv-init`. If a user has elected to export `PATH`, the changes rbenv
makes to the variable will continue to be reflected in the exported
variable.

Other change:
* Remove hard-coded version number in bare invocation test.

### DIFF
--- a/test/rbenv.bats
+++ b/test/rbenv.bats
@@ -5,7 +5,7 @@ load test_helper
 @test "blank invocation" {
   run rbenv
   assert_success
-  assert [ "${lines[0]}" = "$(rbenv---version)" ]
+  assert [ "${lines[0]}" = "rbenv 0.4.0" ]
 }
 
 @test "invalid command" {
--- a/test/rbenv.bats
+++ b/test/rbenv.bats
@@ -5,7 +5,7 @@ load test_helper
 @test "blank invocation" {
   run rbenv
   assert_success
-  assert [ "${lines[0]}" = "rbenv 0.4.0" ]
+  assert [ "${lines[0]}" = "$(rbenv---version)" ]
 }
 
 @test "invalid command" {
--- a/README.md
+++ b/README.md
@@ -165,7 +165,7 @@ easy to fork and contribute any changes back upstream.
    command-line utility.
 
     ~~~ sh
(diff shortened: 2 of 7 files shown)


mislav:
So, modified `PATH` value will be available in sub-processes as well without the `export` clause? Curious. I always though you need `export` to make the value propagate down to sub-processes as well.

test/rbenv.bats
```diff
@@ -5,7 +5,7 @@ load test_helper
 @test "blank invocation" {
   run rbenv
   assert_success
-  assert [ "${lines[0]}" = "rbenv 0.4.0" ]
```
mislav: This change doesn't belong here

geoffharcourt: Without any changes in this spec, removing the `export`s results in this line returning the rbenv version including the Git SHA. I'm not sure how to resolve this.

geoffharcourt:
If a user wants to `export` the `PATH`, they will already be exporting it in their shell configuration, further `export`s aren't needed. Mutating a previously `export`ed variable results in the changes continuing to be reflected for sub-processes.

geoffharcourt:
In my testing it wasn't necessary to `export` the `PATH` for it to be passed on to processes, but as it stands `rbenv init` is taking that decision out of the user's hands.

mislav:
Thanks for the patch and for your patience, this will be a "wontfix". Most of the changes you made to remove `export` were internal in rbenv and test code, and those exports don't affect the user in any way and can remain as they are right now.

The only `export` that affects the user's shell is the one outputted from `rbenv init -` to be eval'd in an interactive shell. You said that we take the decision to export the PATH variable away from the user, and you have a good point there. But, we *want* to export PATH after changes, because if the user doesn't do it, then rbenv is going to be mis-configured in sub-processes (the `shims` directory won't be in PATH) which might lead to hard-to-debug programs.

I think most people typically export PATH in their shell, and won't mind if we follow that practice too. I can't see how this could be a problem for users. The reason I'm reluctant to merge this patch is that I want to release rbenv 1.0 soon and I'm wary that this change could complicate things for some users. I'd rather leave things as they are, unless we are causing some users serious headache.

PR #819, opened 2015-11-12
tests: fix path_without to handle /bin properly

blueyed:
I was seeing /bin being left.  This patch fixes it by anchoring the search
pattern with ':' on both sides.

Also affects pyenv.


mislav:
Thanks!

PR #821, opened 2015-11-13
add completion block for rbenv-help

jasonkarns:
I'm not sure about the proper execution of `rbenv-commands`.

Throughout the codebase I've seen `rbenv commands`, `rbenv-commands`, `command rbenv commands`, and `exec rbenv commands`


mislav:
Nice! `rbenv-commands` is correct, since it will be faster to execute it, and `RBENV_ROOT` is guaranteed to be initialized at this point.

PR #822, opened 2015-11-16
Add --help and tab completion to rbenv-versions

jasonkarns:
- adds --help option for full help
- adds tab completion (--bare, --skip-aliases)

Questions:
- should --help itself be tab completed?
- should --help option be listed in usage?

### DIFF
--- a/libexec/rbenv-completions
+++ b/libexec/rbenv-completions
@@ -10,6 +10,11 @@ if [ -z "$COMMAND" ]; then
   exit 1
 fi
 
+# Provide rbenv completions
+if [ "$COMMAND" = "--complete" ]; then
+  exec rbenv-commands
+fi
+
 COMMAND_PATH="$(command -v "rbenv-$COMMAND" || command -v "rbenv-sh-$COMMAND")"
 
 # --help is provided automatically
--- a/libexec/rbenv-versions
+++ b/libexec/rbenv-versions
@@ -9,8 +9,13 @@ set -e
 
 unset bare
 unset skip_aliases
+# Provide rbenv completions
 for arg; do
   case "$arg" in
+  --complete )
+    echo --bare
+    echo --skip-aliases
+    exit ;;
   --bare ) bare=1 ;;
   --skip-aliases ) skip_aliases=1 ;;
   * )
--- a/libexec/rbenv-help
+++ b/libexec/rbenv-help
@@ -18,8 +18,7 @@ set -e
 # Provide rbenv completions
 if [ "$1" = "--complete" ]; then
   echo --usage
-  rbenv-commands
-  exit
+  exec rbenv-commands
 fi
 
 command_path() {
--- a/libexec/rbenv-init
+++ b/libexec/rbenv-init
@@ -5,6 +5,17 @@
 set -e
 [ -n "$RBENV_DEBUG" ] && set -x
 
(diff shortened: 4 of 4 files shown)


jasonkarns:
Given answers to the above questions, I'd like to extend this PR to make the tab completion for other subcommands consistent.

libexec/rbenv-versions
```diff
@@ -11,7 +11,13 @@ unset bare
 unset skip_aliases
 for arg; do
   case "$arg" in
+  --complete )
+    echo --bare
+    echo --help
```
mislav: Before you add `--help` to every command, we should maybe consider solving that on another level. Right now the preferred way to get help was `rbenv help <command>`. Maybe we could change the router to make `rbenv <command> --help` do the same thing? That would save us from having to edit every subcommand's argument parsing.

libexec/rbenv-versions
```diff
@@ -11,7 +11,13 @@ unset bare
 unset skip_aliases
 for arg; do
   case "$arg" in
+  --complete )
```
mislav: When you add this, you also have to add this line somewhere:

```
# provide rbenv completions
```

It's not intuitive, but this magic comment "marks" the command as being completion-aware.

libexec/rbenv-versions
jasonkarns: I like this idea. I'm mainly modeling this on git, which supports both `git help <command>` and `git <command> --help` (and the fact that some plugins/commands already had --help but others didn't.)

jasonkarns: We could also update the tab completion portion to always echo `--help` as well (before invoking the subcommand's `--complete`).

libexec/rbenv-versions
jasonkarns: right, not sure how I missed that in this one. I definitely had it at one point, because I verified it was working :/

libexec/rbenv-versions
mislav: I guess so, yeah. I don't find that to be extremely valuable, but for the sake of correctness.

jasonkarns:
With the general `--help` feature merged, I've rebased this branch.

- added completions for `completions`, `init`, and `versions`
- made `help`'s completions consistent with others

After this pr, all commands under libexec have respective completions.

libexec/rbenv-completions
```diff
@@ -10,6 +10,11 @@ if [ -z "$COMMAND" ]; then
   exit 1
 fi
 
+# Provide rbenv completions
+if [ "$COMMAND" = "--complete" ]; then
+  exec rbenv-commands
```
mislav: Do you see a value in providing completion support for `rbenv-completion`? After all, people shouldn't generally use this command directly in their interactive shell.

mislav:
Generally looks :

libexec/rbenv-completions
jasonkarns: I'm so-so on it. Most people probably don't, but I use it a lot when monkeying with rbenv/nodenv. So marginal usefulness + consistency + no downsides = enough to convince me. :)

jasonkarns:
I was more unconvinced on the inclusion of the `-` in rbenv-init. But since it's virtually a required argument, I decided to keep it in. :shrug:

PR #824, opened 2015-11-17
Handle --help flag for subcommands

jasonkarns:
First attempt at handling --help uniformly across all subcommands.

If a subcommand is provided and exists and its first subarg is `-h` or `--help`, intercept the call and exec rbenv-help for the subcommand instead.

Additionally, for tab completion, include `--help` in the completions for a subcommand

Prompted by https://github.com/sstephenson/rbenv/pull/822#discussion_r45077235

### DIFF
--- a/libexec/rbenv
+++ b/libexec/rbenv
@@ -97,7 +97,7 @@ case "$command" in
   } >&2
   exit 1
   ;;
--v )
+-v | --version )
   exec rbenv---version
   ;;
 -h | --help )
--- a/libexec/rbenv
+++ b/libexec/rbenv
@@ -111,6 +111,10 @@ case "$command" in
   fi
 
   shift 1
-  exec "$command_path" "$@"
+  if [ "$1" = --help ]; then
+    exec rbenv-help "$command"
+  else
+    exec "$command_path" "$@"
+  fi
   ;;
 esac
--- a/libexec/rbenv-completions
+++ b/libexec/rbenv-completions
@@ -11,6 +11,10 @@ if [ -z "$COMMAND" ]; then
 fi
 
 COMMAND_PATH="$(command -v "rbenv-$COMMAND" || command -v "rbenv-sh-$COMMAND")"
+
+# --help is provided automatically
+echo --help
+
 if grep -iE "^([#%]|--|//) provide rbenv completions" "$COMMAND_PATH" >/dev/null; then
   shift
   exec "$COMMAND_PATH" --complete "$@"
--- a/test/completions.bats
+++ b/test/completions.bats
@@ -13,7 +13,7 @@ create_command() {
   create_command "rbenv-hello" "#!$BASH
     echo hello"
   run rbenv-completions hello
-  assert_success ""
+  assert_success "--help"
 }
 
 @test "command with completion support" {
@@ -25,7 +25,11 @@ else
   exit 1
 fi"
   run rbenv-completions hello
-  assert_success "hello"
+  assert_success
+  assert_output <<OUT
+--help
+hello
+OUT
 }
 
 @test "forwards extra arguments" {
@@ -40,6 +44,7 @@ fi"
   run rbenv-completions hello happy world
   assert_success
   assert_output <<OUT
+--help
 happy
 world
 OUT
--- a/test/exec.bats
+++ b/test/exec.bats
@@ -35,6 +35,7 @@ create_executable() {
   run rbenv-completions exec
   assert_success
   assert_output <<OUT
+--help
 rake
 ruby
 OUT


libexec/rbenv
```diff
@@ -91,20 +93,21 @@ shopt -u nullglob
 
 command="$1"
 case "$command" in
-"" | "-h" | "--help" )
-  echo -e "$(rbenv---version)\n$(rbenv-help)" >&2
```
jasonkarns: shouldn't a blank invocation be an error? -h/--help should print help output with a clean exit. But I think blank invocation should exit 1 after printing help, since it's not a valid invocation. I'm updating the test to reflect this, but wanted to call this out specifically in case it should be reverted

mislav: Sorry, I have worked few days ago on how I want blank invocation to behave, but I have forgot to commit and pushed. I've done this now on master, so you should rebase. Sorry for the conflict.

libexec/rbenv
```diff
@@ -91,20 +93,21 @@ shopt -u nullglob
...
+    exit ;;
+  -v | --version )
+    exec rbenv---version ;;
+  * )
+    command_path="$(command -v "rbenv-$command" || true)"
+    [ -n "$command_path" ] || fail "rbenv: no such command \`$command'"
+
+    shift 1
+    case "$1" in
+      -h | --help ) exec rbenv-help "$command" ;;
```
mislav: Unfortunately we can't support `-h`, because some subcommand might use that as a flag to mean something different, such as short of `--human` or `--head` or whatever. The long flag `--help` is OK because it's specific enough.

mislav:
There are a lot of unrelated changes in this PR, such as indentation and style tweaks to unrelated code.

Please strive for minimal diff so it's easier for me to review this. Minimal diff PRs also have a much higher chance to get merged :wink: thanks!

jasonkarns:
trimmed down, only --help is accepted (not -h). 
guard added for completions that the subcommand is executable and --help is automatically added to completion output for subcommands

libexec/rbenv-completions
```diff
@@ -11,6 +11,11 @@ if [ -z "$COMMAND" ]; then
 fi
 
 COMMAND_PATH="$(command -v "rbenv-$COMMAND" || command -v "rbenv-sh-$COMMAND")"
+[ -x "$COMMAND_PATH" ] || exit 1
```
mislav: Why this check

libexec/rbenv
```diff
@@ -111,6 +111,9 @@ case "$command" in
   fi
 
   shift 1
-  exec "$command_path" "$@"
+  case "$1" in
```
mislav: Simple `if/else` will suffice here.

libexec/rbenv
```diff
@@ -97,7 +97,7 @@ case "$command" in
   } >&2
   exit 1
   ;;
--v )
+-v | --version )
```
mislav: This wasn't strictly necessary, right?

mislav:
Looking sooooo much better! :+1:

libexec/rbenv
jasonkarns: nope: from the commit message:
> Without it, it falls back to `--version` being executed as a subcommand,
which works but isn't obvious

mislav: So this is for flags parsing readability, not helping actual functionality?

mislav: I'm OK with that.

libexec/rbenv
jasonkarns: right. leftover from when I was accepting `-h`

libexec/rbenv-completions
jasonkarns: `command -v` doesn't actually care if the file is executable or not
it's technically possible to invoke `rbenv completions foo`, assuming `rbenv-foo` exists but isn't executable, and have the completions be attempted. It would then fail at the `exec` line since `rbenv-foo` isn't executable

However, with this PR, we're now emitting "--help" prior to `exec $cmd --complete` so it would echo "--help" and *then* error at the exec line.

I suppose it's overly cautious, though. rbenv-completions with a non-executable subcommand still fails the same way as before, it just now echos "--help" before doing so. I thought that was a bit misleading, so I added the guard. I suppose the odds of someone actually having an `rbenv-{thing}` that exists in path somewhere but not being executable are pretty infinitesimal.

yanking

libexec/rbenv
jasonkarns: yep, purely readability. (and short circuits all the rest of the code in the `*)` case.

jasonkarns:
cleaned up

mislav:
We distilled this PR from `+62 -47` down to `+18 −4`. Thanks for bearing with me!

jasonkarns:
great! those other changes will be forthcoming as separate PRs ;)

PR #826, opened 2015-11-20
Extract `fail` helper function

jasonkarns:
fail helper prints message to stderr and exits 1

### DIFF
--- a/libexec/rbenv
+++ b/libexec/rbenv
@@ -12,21 +12,24 @@ if [ -n "$RBENV_DEBUG" ]; then
   set -x
 fi
 
+abort() {
+  { if [ "$#" -eq 0 ]; then cat -
+    else echo "rbenv: $*"
+    fi
+  } >&2
+  exit 1
+}
+
 if enable -f "${BASH_SOURCE%/*}"/../libexec/rbenv-realpath.dylib realpath 2>/dev/null; then
   abs_dirname() {
     local path="$(realpath "$1")"
     echo "${path%/*}"
   }
 else
-  if [ -n "$RBENV_NATIVE_EXT" ]; then
-    echo "rbenv: failed to load \`realpath' builtin" >&2
-    exit 1
-  fi
+  [ -z "$RBENV_NATIVE_EXT" ] || abort "failed to load \`realpath' builtin"
+
 READLINK=$(type -p greadlink readlink | head -1)
-if [ -z "$READLINK" ]; then
-  echo "rbenv: cannot find readlink - are you missing GNU coreutils?" >&2
-  exit 1
-fi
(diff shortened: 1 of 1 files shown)


libexec/rbenv
```diff
@@ -12,21 +12,25 @@ if [ -n "$RBENV_DEBUG" ]; then
   set -x
 fi
 
+fail() {
+  {
```
mislav: I usually start the block right on the opening bracket line:

```sh
{ if [ ... ]; ....
  ...
}
```

libexec/rbenv
```diff
@@ -12,21 +12,25 @@ if [ -n "$RBENV_DEBUG" ]; then
...
+    if [ "$#" -eq 0 ]; then cat -
+    else echo -e "$@"
```
mislav: Why `-e`?

mislav:
Thanks for the suggestion. I see how this encapsulates a pattern we repeat in the same file a lot; I'm just not sure whether it's really worthwhile to DRY this up. It's really not that hard to redirect to stderr and `exit 1`. But I'll think about this :+1:

libexec/rbenv
```diff
@@ -12,21 +12,25 @@ if [ -n "$RBENV_DEBUG" ]; then
   set -x
 fi
 
+fail() {
```
mislav: Perhaps we could call it `abort`?

libexec/rbenv
```diff
@@ -92,10 +93,7 @@ shopt -u nullglob
 command="$1"
 case "$command" in
 "" )
-  { rbenv---version
-    rbenv-help
-  } >&2
-  exit 1
+  { rbenv---version; rbenv-help; } | fail
```
mislav: I would split this into separate lines like before

libexec/rbenv
mislav: How about if a string argument was given, we prefix it with `rbenv: ` always? That way we don't have to repeat `rbenv: ` in each message in this file.

libexec/rbenv
jasonkarns: :+1:

libexec/rbenv
jasonkarns: that breaks vim's auto-formatting/indentation for SH files.

```
abort() {
  { if [ "$#" -eq 0 ]; then cat -
  else echo -e "$@"
  fi
} >&2
exit 1
}
```

But can switch if still desired

mislav: Yeah, please switch to match the currently prevalent style in rbenv and ruby-build

jasonkarns:
i think i addressed everything

mislav:
Thanks. How about squashing this

jasonkarns:
squashed

mislav:
Thanks!

PR #827, opened 2015-11-20
Fix leftover .ruby-version file from exec.bats

jasonkarns:
This new test was creating an (intentionally invalid) .ruby-version file
in current working directory; typically the rbenv project dir.
Immediately after test runs, I had a leftover .ruby-version file.

The version-file tests create and cd into the RBENV_TEST_DIR as part of
setup(). I'm using the same directory for this test fix, but am only
using it for this particular test. None of the other exec tests seem to
need to be in a temp test dir, so no use putting it in setup().

### DIFF
--- a/test/exec.bats
+++ b/test/exec.bats
@@ -21,6 +21,8 @@ create_executable() {
 }
 
 @test "fails with invalid version set from file" {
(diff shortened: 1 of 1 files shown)


mislav:
Nice catch!

PR #828, opened 2015-11-20
indentation

jasonkarns:
This just fixes the indentation (used vim's out-of-the-box shell formatter)

There is only one non-whitespace change in the PR: switching `$(pwd)` to `$PWD`
This matches a similar change that was made to rbenv-hooks
(7026e529c726648a30af0666088da541245d6422).
rbenv-version-file and rbenv-versions have use `$PWD` since the beginning

https://github.com/sstephenson/rbenv/commit/7026e529c726648a30af0666088da541245d6422#diff-a2b5e4e0b5ebc86ede5c3f80181712a4L38
https://github.com/sstephenson/rbenv/commit/6c1fb9ffd062ff04607d2e0f486067eaf6e48d1e#diff-0b6c3d2c0fa07263937b4ec78b712059R17
https://github.com/sstephenson/rbenv/commit/e80886e9bed1fb51d6df7d37725537fae875e3f6#diff-9befc8fccb1c8d66eaa69fae65f613a5R42

@mislav, this is my last pr, i swear :)

### DIFF
--- a/libexec/rbenv
+++ b/libexec/rbenv
@@ -36,7 +36,7 @@ else
   }
 
   abs_dirname() {
-    local cwd="$(pwd)"
+    local cwd="$PWD"
     local path="$1"
 
     while [ -n "$path" ]; do
@@ -58,10 +58,10 @@ fi
 export RBENV_ROOT
 
 if [ -z "${RBENV_DIR}" ]; then
-  RBENV_DIR="$(pwd)"
+  RBENV_DIR="$PWD"
 else
   cd "$RBENV_DIR" 2>/dev/null || abort "cannot change working directory to \`$RBENV_DIR'"
-  RBENV_DIR="$(pwd)"
+  RBENV_DIR="$PWD"
   cd "$OLDPWD"
 fi
 export RBENV_DIR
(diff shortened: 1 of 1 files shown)


libexec/rbenv
```diff
@@ -91,26 +91,26 @@ shopt -u nullglob
 
 command="$1"
 case "$command" in
-"" )
```
mislav: I like case statements to be indented to the same level as opening `case` keyword.

jasonkarns: Ya just had to have an exception so that the vim autoformatter couldn't be used, didn't you. ;)

updated

related: if you use vim, do you have another formatter config that I could use? I couldn't care less about the actual indentation form, as long as it can be automated.

mislav: No, I use mostly defaults in vim, but I don't use autoformatter for either bash or Ruby since it's often wrong.

jasonkarns:
Thanks to this discussion, I discovered that `sh.vim` (from vim-scripts) has made the case indentation configurable. It now defaults to case-labels at the same level as `case`. To increase the label indentation: `let g:sh_indent_case_labels=1`

According to the [readme on github](https://github.com/vim-scripts/sh.vim), this is the only difference between sh.vim and the built-in formatter.

PR #829, opened 2015-11-21
Detect `RBENV_DIR` properly from file argument to `ruby`

yyuu:
I believe there are 2 weird behaviour relating to `RBENV_DIR`.  This should fix both 2.
1. file argument does not follow symlink (yyuu/pyenv#404)
2. file argument works weirdly if first argument doesn't contain `/` (yyuu/pyenv#479)

Both can be reproduced with the following setup.

``` sh
$ mkdir -p ./foo
$ echo 2.2.3 > ./foo/.ruby-version
$ echo 'puts(RUBY_DESCRIPTION)' > ./foo/test.rb
$ mkdir -p ./foo/bar
$ echo jruby-9.0.0.0 > ./foo/bar/.ruby-version
$ ln -sf ${PWD}/foo/test.rb ./foo/bar/
```

If the first file argument to `ruby` interpreter is a symlink, it'd be better to follow the link.

``` sh
$ ruby ./foo/test.rb
ruby 2.2.3p173 (2015-08-18 revision 51636) [x86_64-linux]
$ ruby ./foo/bar/test.rb
jruby 9.0.0.0 (2.2.2) 2015-07-21 e10ec96 OpenJDK 64-Bit Server VM 25.72-b05 on 1.8.0_72-internal-b05 +jit [linux-amd64]
```

`RBENV_DIR` just not working if the first file argument doesn't contain `/`.

``` sh
$ cd ./foo
$ ruby test.rb
ruby 2.2.3p173 (2015-08-18 revision 51636) [x86_64-linux]
$ ruby test.rb ./bar/test.rb
jruby 9.0.0.0 (2.2.2) 2015-07-21 e10ec96 OpenJDK 64-Bit Server VM 25.72-b05 on 1.8.0_72-internal-b05 +jit [linux-amd64] # <-- this would be better to be 2.2.3
```

### DIFF
--- a/libexec/rbenv
+++ b/libexec/rbenv
@@ -58,7 +58,13 @@ fi
 export RBENV_ROOT
 
 if [ -z "${RBENV_DIR}" ]; then
-  RBENV_DIR="$PWD"
+  # Transfer RBENV_FILE_ARG (from shims) into RBENV_DIR.
+  if [ -n "${RBENV_FILE_ARG}" ]; then
+    RBENV_DIR="$(abs_dirname "$(dirname "${RBENV_FILE_ARG}")/${RBENV_FILE_ARG##*/}")"
+    unset RBENV_FILE_ARG
+  else
+    RBENV_DIR="$PWD"
+  fi
 else
   cd "$RBENV_DIR" 2>/dev/null || abort "cannot change working directory to \`$RBENV_DIR'"
   RBENV_DIR="$PWD"
--- a/libexec/rbenv-rehash
+++ b/libexec/rbenv-rehash
@@ -50,9 +50,9 @@ if [ "\$program" = "ruby" ]; then
   for arg; do
     case "\$arg" in
     -e* | -- ) break ;;
-    */* )
+    * )
       if [ -f "\$arg" ]; then
-        export RBENV_DIR="\${arg%/*}"
+        export RBENV_FILE_ARG="\$arg"
         break
       fi
       ;;
--- a/test/rbenv.bats
+++ b/test/rbenv.bats
@@ -46,6 +46,41 @@ load test_helper
   assert_output "rbenv: cannot change working directory to \`$dir'"
 }
 
+@test "RBENV_DIR should have higher precedence than RBENV_FILE_ARG" {
+  dir="${BATS_TMPDIR}/myproject"
+  mkdir -p "$dir"
+  touch "$dir/test.rb"
+  RBENV_DIR="$dir" RBENV_FILE_ARG="$dir/test.rb" run rbenv echo RBENV_DIR
+  assert_output "$dir"
+}
+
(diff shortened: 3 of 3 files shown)


libexec/rbenv
```diff
@@ -58,10 +58,16 @@ fi
 export RBENV_ROOT
 
 if [ -z "${RBENV_DIR}" ]; then
-  RBENV_DIR="$(pwd)"
+  # Transfer RBENV_FILE_ARG (from shims) into RBENV_DIR.
+  if [ -n "${RBENV_FILE_ARG}" ]; then
+    RBENV_DIR="$(abs_dirname "$(dirname "${RBENV_FILE_ARG}")/${RBENV_FILE_ARG##*/}")"
```
mislav: Why to `"$(dirname "${RBENV_FILE_ARG}")/${RBENV_FILE_ARG##*/}"` when that seems equivalent of `"$RBENV_FILE_ARG"`?

yyuu: It is to handle files without slash in their name properly. `abs_dirname` requires argument to have at least one slash in its name. Because `RBENV_FILE_ARG` might not contain any slash in it, I need to use `dirname` instead of string manipulation for it.

jasonkarns: I've done `abs_dirname "${1%/}/"` for similar effect.

yyuu: @jasonkarns Interesting. Though it is not applicable for this case since `RBENV_FILE_ARG` is a file, not a directory.

yyuu:
Resolved conflict.

libexec/rbenv
jasonkarns: Ah, I misunderstood the intent of the dirname. I thought this could be simplified to simply prepend `./` to any value that didn't already start with a `/`

jasonkarns: Seems like this is a deficiency in abs_dirname, though. Would it be better to fix abs_dirname to still do the right thing when passed just a simple filename?

yyuu: The intention of the usage of `dirname` is I just wanted to do thing in one-liner. Prepending `./` would just work.

Fixing `abs_dirname` would be the best. Though there is a copy of `abs_dirname` in `ruby-build`. It'd be better to fix both if we modified it.

jasonkarns: It's also been copied to some plugins

mislav:
Thank you for the implementation, explanation, and the ensuing discussion, but I remained unconvinced that this is a change I want in rbenv, and in the interest of keeping rbenv stable, I'm going to close this. Furthermore, I stopped resolving symlinks in most places within rbenv: https://github.com/rbenv/rbenv/pull/1428

I believe that if you call `ruby /path/to/file.rb`, the `/path/to/.ruby-version` should be respected and the actual destination of the `/path/to/file.rb` symlink shouldn't really matter.

PR #832, opened 2015-12-01
Update README.md for Arch Linux install

jonathanpa:
Add section to install rbenv and ruby-build from AUR packages on Arch Linux.


mislav:
Sorry, we don't currently maintain a list of package managers installation instructions for platforms we're unfamiliar with, except Homebrew on OS X which is what we the maintainers use.

Could you instead add this information to the wiki? https://github.com/rbenv/rbenv/wiki/Installation-with-package-managers

PR #836, opened 2015-12-12
Remove leading `:`

eagletmt:
Currently, `rbenv which` returns the current directory's command when RBENV_VERSION=system.

```
% ls -l ls
-rwxr-xr-x 1 eagletmt eagletmt 0 2015-12-12 15:36 ls
% which ls
/usr/bin/ls
% rbenv which ls
./ls
```

This patch fixes the issue.

### DIFF
--- a/libexec/rbenv-which
+++ b/libexec/rbenv-which
@@ -23,7 +23,8 @@ remove_from_path() {
     path_before="$result"
     result="${result//:$path_to_remove:/:}"
   done
-  echo "${result%:}"
+  result="${result%:}"
+  echo "${result#:}"
 }
 
 RBENV_COMMAND="$1"


hsbt:
@mislav It seems to fix regression. We should merge this.

mislav:
Thanks!

mislav:
In the future, adding a test would be great, but this diff looked safe enough to merge right away.

mislav:
This originally regressed in e4cbf04592cfcd9d244fb83284e14524c8ba0377. Pinging @blueyed for awareness that they should pull the fix into pyenv as well. I've added a test in 90373d78b91484b738d57f2bb54c6b417c95e182

This bug did not affect v0.4.0 (last stable release).

blueyed:
@mislav 
Thanks!  Will be merged through https://github.com/yyuu/pyenv/pull/504 probably.

PR #838, opened 2015-12-14
Fix error about symbolic link

mengbo:
Fix error about symbolic link https://github.com/yyuu/pyenv/pull/502

### DIFF
--- a/libexec/rbenv
+++ b/libexec/rbenv
@@ -40,7 +40,7 @@ else
     local path="$1"
 
     while [ -n "$path" ]; do
-      cd "${path%/*}"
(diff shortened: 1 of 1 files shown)


mislav:
Thanks for the fix! Could this come with a test that fails without a fix? Otherwise it's hard to tell whether this fixes anything.

blueyed:
I came up with another fix in https://github.com/rbenv/rbenv/pull/868 - which might be more performant?!

But it's also missing a test.

blueyed:
Where should the test go?

mislav:
Closing as wontfix since I'm not sure what this actually fixes.

PR #842, opened 2015-12-24
Check if shims is already in path before adding it

bradynpoulsen:


bradynpoulsen:
I did a new pull request because I did a bad rebase

mislav:
We had this feature but I had to revert (e2173df4aa91c8d365ca1596fb857fcac9fdd787) because it bit too many people who had duplicated PATH configuration due to starting terminal multiplexer such as tmux. These users needed the shims path to get prepended to the PATH again with high precedence, but because of non-duplication logic, the shims path would stay in PATH with low precedence while other system paths were repeated on top.

So, for the sake of simplicity and not being overly clever in code, we're not gonna do this.

BTW you don't have to reopen a new PR when you mess one up. You can force-push to the same branch to refresh your PR after a bad rebase.

bradynpoulsen:
@mislav Makes sense. Thanks.

For now I'm just doing `which rbenv > /dev/null && [[ ":$PATH:" != *":$(rbenv root)/shims:"* ]] && eval "$(rbenv init -)"`

Thanks

mislav:
@bradynpoulsen In which case do you get duplicate paths? Also tmux?

bradynpoulsen:
I've never really used tmux before, but just to figure it out I adjusted it to
```sh
if which rbenv > /dev/null; then
    [[ "$PATH:" != "$(rbenv root)/shims:"* ]] && eval "$(rbenv init -)"
    export PATH="${PATH//:$(rbenv root)\/shims:/}"
fi
```

bradynpoulsen:
I use a `pathAdd <newPath>` function to conditionally add paths to the `PATH`. I've adjusted it to do the following
```sh
function pathAdd() {
    if [[ -n "$1" ]]; then
        [[ ":$PATH:" == *":/sbin:"*":$1"* ]] && PATH=`echo $PATH | sed -e "s/:${1//\//\\\/}//"`
        if [[ -d "$1" ]] && [[ ":$PATH:" != *":$1:"* ]]; then
            PATH="$1:$PATH"
        fi
    fi
}
```

Issue #631, opened 2014-10-06
How to unset global Ruby version?


grepwood:
I've a script that uninstalls Ruby installed via rbenv and makes sure it is uninstalled. `rbenv versions` will complain that the global version of Ruby is not installed after it's uninstalled. Is there any way to unset the global?

Issue #362, opened 2013-03-16
"rbenv global system" doesn't unset global Ruby when there's no system Ruby installed


fj:
If you have no actual system Ruby, then rbenv can't unset whatever Ruby you're currently using if you want to fall back to no Ruby at all:

```
$ rbenv global system
rbenv: ruby: command not found

The `ruby' command exists in these Ruby versions:
  2.0.0-p0
```

This workaround of deleting the `~/.rbenv/version` file will work though:

```
$ rbenv version
2.0.0-p0 (set by /home/jxf/.rbenv/version)

$ rm -rf ~/.rbenv/version

$ rbenv version
system (set by /home/jxf/.rbenv/version)
```

PR #848, opened 2015-12-29
Add `--unset` to rbenv-global

jasonkarns:
Support unsetting global ruby version. Makes `rbenv global` consistent with
`rbenv local` and `rbenv shell`

### DIFF
--- a/libexec/rbenv-global
+++ b/libexec/rbenv-global
@@ -17,6 +17,7 @@ set -e
 
 # Provide rbenv completions
 if [ "$1" = "--complete" ]; then
+  echo --unset
   echo system
   exec rbenv-versions --bare
 fi
--- a/libexec/rbenv-global
+++ b/libexec/rbenv-global
@@ -22,7 +22,11 @@ if [ "$1" = "--complete" ]; then
 fi
 
 RBENV_VERSION="$1"
-RBENV_VERSION_FILE="${RBENV_ROOT}/version"
+RBENV_VERSION_FILE="${RBENV_ROOT:?}/version"
+
+if [ "$1" = "--unset" ]; then
+  exec rm -f "$RBENV_VERSION_FILE" "${RBENV_ROOT:?}/global" "${RBENV_ROOT:?}/default"
+fi
 
 if [ -n "$RBENV_VERSION" ]; then
   rbenv-version-file-write "$RBENV_VERSION_FILE" "$RBENV_VERSION"
--- a/test/global.bats
+++ b/test/global.bats
@@ -3,7 +3,7 @@
 load test_helper
 
 @test "default" {
-  run rbenv global
+  run rbenv-global
   assert_success
   assert_output "system"
 }
@@ -28,4 +28,16 @@ load test_helper
   mkdir -p "$RBENV_ROOT"
   run rbenv-global "1.2.3"
   assert_failure "rbenv: version \`1.2.3' not installed"
+}
+
+@test "unset (remove) RBENV_ROOT/version" {
+  mkdir -p "$RBENV_ROOT"
+  echo "1.2.3" > "$RBENV_ROOT/version"
+
+  run rbenv-global --unset
+  assert_success
+
+  assert [ ! -e $RBENV_ROOT/version ]
+  run rbenv-global
+  assert_output "system"
 }


mislav:
I'm not sure if I want people to be able to unset the global version. One should change it to "system" instead. Maybe `--unset` should be an equivalent of that? But then it's a misnomer, because you can't truly unset a global rbenv version— it just falls back to system if the file is missing.

Related: #631 #362

jasonkarns:
I know there isn't a technical difference between removing the global version file and setting `rbenv global system`. Do you think there is a semantic difference?

When my goal is to revert to the system ruby, for some reason my brain doesn't think `rbenv {global,local,shell} system`, I think `--unset`. When I unset local/shell versions, it's because I want to remove the more-specific setting and just fall back to global. Similarly, I mentally think of the system ruby as another level of fallback behind `global`. (`shell` -> `local` -> `global` -> system) So when I want the global version to be system, I don't think in terms of 'setting' the global version to system, but rather clearing the global version and letting it fall back to system. As if I'm (sorta) removing rbenv from the equation. In this view, there is a (hair-splitting) semantic difference between removing the global file and setting it to `system`. I just think in terms of *clearing* the global version, not setting it to `system`.

I think this is why I find the lack of global-unset a little incongruent.

jasonkarns:
FWIW, this change was originally prompted by nodenv. Since node doesn't ship with OS X, I find that many people have the 'system' node installed via homebrew, and use nodenv to manage additional nodes. Since homebrew deps are generally upgraded regularly, one's system node is (generally) fairly recent. Thus, I find myself frequently pegging the global node to a nodenv version for a short time (for a number of shells across a number of directories, so local/shell are not suitable). But then when I'm finished with that I want to clear the global setting, and let it fall back to the homebrew node. `nodenv global system` would work fine, but as I described above, my brain doesn't reach for that for whatever reason.

I would be happy with making `--unset` just do a `global system` as you describe. I still think that having `--unset` available would be nice.

mislav:
Thanks for explaining. I'm still not convinced we need `--unset`. In the case of `shell` and `local`, the `--unset` flag literally deletes a setting. In the case of `global`, it would just change it.

> one's system node is (generally) fairly recent. Thus, I find myself frequently pegging the global node to a nodenv version for a short time

Sounds like you want to temporarily set your global nodenv version to "system" node. I don't see why `nodenv global system` would be counterintuitive here.

Not many people ask for this, therefore I'm not terribly motivated to approve this change yet.

jasonkarns:
> In the case of shell and local, the --unset flag literally deletes a setting. In the case of global, it would just change it.

well.... yeah, I guess. As this PR is currently authored, global --unset deletes a setting, too. It's a perfect mirror of the implementation for shell/local --unset. As long as the global version is set via a file, it's possible to delete the version file just as one could delete a local version file. Having a way to delete the file via the cli would be convenient and makes the shell/local/global subcommands all consistent.

> I don't see why `nodenv global system` would be counterintuitive here.

I'm not necessarily saying it's counter-intuitive for others, just that I personally didn't even think to try it. TBH, until you mentioned it, I never even realized that `system` was a valid value that could be set. I just assumed without looking into it deeply, that 'system' was simply a helpful display-only message for when rbenv didn't set a ruby version. I had no idea that rbenv itself could actually be set to `system`.

aried3r:
Hey! I think I am facing this issue as well. Maybe it's the same use case.

I had Ruby 2.5.1 installed on an Ubuntu installation that had no Rubies itself and set it to be the global version. When I uninstalled 2.5.1 I get this output from `rbenv versions`

```console
$ rbenv versions
rbenv: version `2.5.1' is not installed (set by /usr/local/rbenv/version)
Warning: no Ruby detected on the system
```

While I expected the warning, I didn't expect the first line which made me google and find out if there's a better way than deleting the `version` file (or emptying it?) which led me to this issue.

> Not many people ask for this, therefore I'm not terribly motivated to approve this change yet.

Well, once every 3 years it seems ;)

Do we still not want this feature? If so, is deleting or emptying the `version` file preferred?

mislav:
> If so, is deleting or emptying the `version` file preferred?

Yes, that is preferred. Closing since there was not much demand for this, and the feature set of rbenv is basically frozen for the time being. Thank you @jasonkarns for the idea!

PR #849, opened 2015-12-29
Remove all handling/support of .rbenv-version files

### LINKED ISSUE #846: Remove all handling/support of .rbenv-version files
Rbenv migrated from rbenv-version to ruby-version files a long time ago. (late 2012, I believe)

There is plenty of code that handles rbenv-version as a fallback that could be cleaned up. Now that 1.0 has landed, I propose we remove all the handling and references to rbenv-version files.

I'm happy to submit a PR for this if there is agreement on this direction.

jasonkarns:
closes #846

Rbenv migrated from rbenv-version to ruby-version files in late 2012.

It's been nearly 3 years now and 1.0 has been released. This PR removes support for the legacy file: .rbenv-version

### DIFF
--- a/README.md
+++ b/README.md
@@ -374,11 +374,6 @@ configured local version. You can also unset the local version:
 
     $ rbenv local --unset
 
-Previous versions of rbenv stored local version specifications in a
-file named `.rbenv-version`. For backwards compatibility, rbenv will
-read a local version specified in an `.rbenv-version` file, but a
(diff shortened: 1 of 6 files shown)


jasonkarns:
Of special note is the first commit (3c9674453fe51b40acef22141ad477d8da86dc7e) which fixes a false-negative test that has likely been testing nothing since `.ruby-version` support was first added.

mislav:
Thanks for the cleanup!

PR #850, opened 2015-12-29
Remove support for legacy global version files

jasonkarns:
In the spirit of #849, I propose we remove the `default` and `global` legacy fallbacks for the global filename (`$RBENV_ROOT/version` has been the preferred filename since late 2011)

`default` was made legacy back in 2011 with 5be66da9f4053137a320058495031d0af6db29b4 (the command was renamed from `rbenv-default` to `rbenv-global`, and so the global file was renamed from `$RBENV_ROOT/default` to `$RBENV_ROOT/global` (the latter taking precedence)

`global` was then made legacy about a month later in Sep 2011 with 0a4ffcd15f6d02f9c707af008d9c2542dc1f864d when the preferred filename was changed to `$RBENV_ROOT/version`.

### DIFF
--- a/libexec/rbenv-global
+++ b/libexec/rbenv-global
@@ -27,8 +27,5 @@ RBENV_VERSION_FILE="${RBENV_ROOT}/version"
 if [ -n "$RBENV_VERSION" ]; then
   rbenv-version-file-write "$RBENV_VERSION_FILE" "$RBENV_VERSION"
 else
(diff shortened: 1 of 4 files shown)


jasonkarns:
Note, this will necessitate a change to #848 (if this is merged, then #848 should be updated to not delete `$RBENV_ROOT/global` and `$RBENV_ROOT/default`)

mislav:
Thanks for the cleanup! :sparkles:

PR #852, opened 2016-01-03
Extract hook helper

jasonkarns:
`create_hook` already existed in `hooks.bats`. The essential body of create_hook was already in use in `exec.bats`, `rehash.bats`, `version-name.bats`, `version-origin.bats` and `which.bats`.

Extracted existing create_hook helper from hooks.bats. Modified it to write the hook to `RBENV_HOOK_PATH` and to accept hook body from stdin.

### DIFF
--- a/test/hooks.bats
+++ b/test/hooks.bats
@@ -2,11 +2,6 @@
 
 load test_helper
 
-create_hook() {
-  mkdir -p "$1/$2"
-  touch "$1/$2/$3"
-}
-
 @test "prints usage help given no argument" {
   run rbenv-hooks
   assert_failure "Usage: rbenv hooks <command>"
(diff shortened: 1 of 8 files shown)


jasonkarns:
I can squash this before merging. Left as is for review

test/hooks.bats
```diff
@@ -15,11 +10,13 @@ create_hook() {
 @test "prints list of hooks" {
   path1="${RBENV_TEST_DIR}/rbenv.d"
   path2="${RBENV_TEST_DIR}/etc/rbenv_hooks"
-  create_hook "$path1" exec "hello.bash"
-  create_hook "$path1" exec "ahoy.bash"
-  create_hook "$path1" exec "invalid.sh"
-  create_hook "$path1" which "boom.bash"
-  create_hook "$path2" exec "bueno.bash"
+  RBENV_HOOK_PATH="$path1"
+  echo | create_hook exec "hello.bash"
```
mislav: Not a fan of `echo | create_hook`. Within `create_hook`, you should be able to check if it has stdin by `[ ! -t 0 ]`.

jasonkarns:
Nice. I was wondering how to check stdin first and use the arg as a fallback (checking number of args wasn't much nicer since we'd want third arg to default to empty string)

Fixed.

test/test_helper.bash
```diff
@@ -129,3 +130,15 @@ path_without() {
   path="${path#:}"
   echo "${path%:}"
 }
+
+create_hook() {
+  local hook_body
+  if [ -t 0 ]; then
+    hook_body="$3"
```
mislav: Could we not support a string argument? I would prefer that creating hooks with content is always the same: by passing stdin using heredoc.

jasonkarns: Sure thing

jasonkarns:
Removed string-argument support for hook body; only accepted now via STDIN

mislav:
Thanks!

Issue #854, opened 2016-01-08
Add `rbenv shell -` to switch back to previous version of Ruby


tenderlove:
Hi,

I think it would be cool if you could do `rbenv shell -` and have `-` mean "the previous version I was using".  Similar to `cd -` and `git checkout -`.  I usually use trunk Ruby installed in to my ruby-build directories, but sometimes I need to switch to stable, then switch back.

I'm not sure if this is the right place to file the ticket.  Also, I'll try to implement this eventually but I don't have the time right now. I just wanted to file a ticket so that I won't forget and in case other people have time.

Thanks!

mislav:
Thanks for the suggestion! The implementation should be trivial: the previous value should simply be saved in a temporary environment variable that doesn't have to be exported like RBENV_VERSION is.

However, what would be the result of running `rbenv shell -` when RBENV_VERSION was previously unset before running `rbenv shell 2.2.4`, for instance?

tenderlove:
> However, what would be the result of running rbenv shell - when RBENV_VERSION was previously unset before running rbenv shell 2.2.4, for instance?

I think we could just do something similar to bash:

```
Last login: Mon Jan 11 11:38:10 on ttys011
[aaron@TC eventmachine (master)]$ cd -
-bash: cd: OLDPWD not set
[aaron@TC eventmachine (master)]$
```

shakycode:
:+1:

mislav:
Merged in 2207c27cb4fce9e7221cd4b1c715b95267538334

PR #857, opened 2016-01-12
WIP add support for rbenv shell -

tenderlove:
This tries to implement `rbenv shell -` as proposed in #854.  However,
adding support seemed to break the "shell change version" test.  I'm not
very good at Bash programming, can someone tell me what is wrong with
what I'm doing?  I'd like to add a bit more functionality to this, but
I'm really just cargo cult programming Bash.

Thank you!

### DIFF
--- a/libexec/rbenv-sh-shell
+++ b/libexec/rbenv-sh-shell
@@ -39,10 +39,32 @@ fi
 if [ "$version" = "--unset" ]; then
   case "$shell" in
   fish )
+    echo "set -e OLD_RBENV_VERSION"
     echo "set -e RBENV_VERSION"
     ;;
   * )
+    echo "unset OLD_RBENV_VERSION"
     echo "unset RBENV_VERSION"
+    ;;
+  esac
+  exit
+fi
+
+if [ "$version" = "-" ]; then
+  if [ -z "$OLD_RBENV_VERSION" ]; then
+    echo "rbenv: OLD_RBENV_VERSION not set" >&2
+    exit 1;
+  fi
+  case "$shell" in
+  fish )
+    rbenv_version=$RBENV_VERSION
+    echo "set -e OLD_RBENV_VERSION \"$rbenv_version\""
+    echo "set -e RBENV_VERSION \"$OLD_RBENV_VERSION\""
+    ;;
+  * )
+    rbenv_version=$RBENV_VERSION
+    echo "export OLD_RBENV_VERSION=\"$rbenv_version\""
+    echo "export RBENV_VERSION=\"$OLD_RBENV_VERSION\""
     ;;
   esac
   exit
@@ -50,14 +72,17 @@ fi
 
 # Make sure the specified version is installed.
 if rbenv-prefix "$version" >/dev/null; then
-  case "$shell" in
-  fish )
-    echo "setenv RBENV_VERSION \"${version}\""
-    ;;
-  * )
-    echo "export RBENV_VERSION=\"${version}\""
-    ;;
-  esac
+  if [ "$version" != "$RBENV_VERSION" ]; then
+    case "$shell" in
+    fish )
+      echo "setenv RBENV_VERSION \"${version}\""
+      ;;
+    * )
+      echo "export OLD_RBENV_VERSION=\"$RBENV_VERSION\""
+      echo "export RBENV_VERSION=\"${version}\""
+      ;;
+    esac
+  fi
 else
   echo "false"
   exit 1
--- a/test/shell.bats
+++ b/test/shell.bats
@@ -22,12 +22,18 @@ load test_helper
 
 @test "shell unset" {
   RBENV_SHELL=bash run rbenv-sh-shell --unset
-  assert_success "unset RBENV_VERSION"
+  assert_output <<OUT
+unset OLD_RBENV_VERSION
+unset RBENV_VERSION
+OUT
 }
 
 @test "shell unset (fish)" {
   RBENV_SHELL=fish run rbenv-sh-shell --unset
-  assert_success "set -e RBENV_VERSION"
+  assert_output <<OUT
+set -e OLD_RBENV_VERSION
+set -e RBENV_VERSION
+OUT
 }
 
 @test "shell change invalid version" {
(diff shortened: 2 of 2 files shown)


tenderlove:
When I run the tests, I get:

```
[aaron@TC .rbenv (dash)]$ bats test/shell.bats 
 ✓ no shell version
 ✓ shell version
 ✓ shell version (fish)
 ✓ shell unset
 ✓ shell unset (fish)
 ✓ shell change invalid version
 ✗ shell change version
   (from function `assert_equal' in file test/test_helper.bash, line 65,
    from function `assert_output' in file test/test_helper.bash, line 74,
    from function `assert_success' in file test/test_helper.bash, line 49,
    in test file test/shell.bats, line 45)
     `assert_success 'export OLD_RBENV_VERSION=""'' failed
   expected: export OLD_RBENV_VERSION=""
   actual:   export OLD_RBENV_VERSION=""
   export RBENV_VERSION="1.2.3"
 ✗ shell change version dash
   (from function `assert_equal' in file test/test_helper.bash, line 65,
    from function `assert_output' in file test/test_helper.bash, line 74,
    from function `assert_success' in file test/test_helper.bash, line 49,
    in test file test/shell.bats, line 53)
     `assert_success 'export RBENV_VERSION="1.2.3"'' failed
   expected: export RBENV_VERSION="1.2.3"
   actual:   export OLD_RBENV_VERSION=""
   export RBENV_VERSION="1.2.3"
 ✓ shell change version (fish)

9 tests, 2 failures
[aaron@TC .rbenv (dash)]$
```

This part is especially confusing to me:

```
     `assert_success 'export OLD_RBENV_VERSION=""'' failed
   expected: export OLD_RBENV_VERSION=""
   actual:   export OLD_RBENV_VERSION=""
```

They look identical, so I'm not sure what I'm doing wrong.

jasonkarns:
the `assert_success` helper asserts a successful return status code (0) as well as asserting that all captured STDOUT matches the expected value. Keyword being *all* output.

```
assert_success 'export OLD_RBENV_VERSION=""'' failed
   expected: export OLD_RBENV_VERSION=""
   actual:   export OLD_RBENV_VERSION=""
```
is actually testing:

```
assert_success 'export OLD_RBENV_VERSION=""'' failed
   expected: export OLD_RBENV_VERSION=""
   actual:   export OLD_RBENV_VERSION=""
   export RBENV_VERSION="1.2.3"
```

I'd merge the two assert_success lines into a single assertion that matches the multiline string:

```
export OLD_RBENV_VERSION=""
export RBENV_VERSION="1.2.3"
```

jasonkarns:
alternatively (and arguably, to me at least, more clear):

``` bash
assert_success
assert_output <<OUT
export OLD_RBENV_VERSION="1.2.3"
export RBENV_VERSION="1.2.4"
OUT
```

tenderlove:
@jasonkarns ahhh, thank you very much.  I think I'll switch to the `assert_output` version.  Thanks!

jasonkarns:
@tenderlove yep! thanks for this feature. I'm excited to have it in rbenv (and soon, nodenv as well)

test/shell.bats
```diff
@@ -42,9 +42,20 @@ SH
 @test "shell change version" {
   mkdir -p "${RBENV_ROOT}/versions/1.2.3"
   RBENV_SHELL=bash run rbenv-sh-shell 1.2.3
+  assert_success 'export OLD_RBENV_VERSION=""'
   assert_success 'export RBENV_VERSION="1.2.3"'
 }
 
+@test "shell change version dash" {
```
jasonkarns: nitpick, but since `dash` is actually a shell that exists, maybe use the literal `-` in the test description. (since the other tests distinguish between bash and fish, dash could theoretically be supported as well)

tenderlove: I will change it.  Thanks!

tenderlove:
I've updated my PR to be a bit better (and have passing tests).  The main issue right now (and I *think* the issue that @mislav was trying to get at in #854) is what to do when nobody has executed `rbenv shell` yet.

For example, I open the terminal and do:

```
Last login: Tue Jan 12 15:44:07 on ttys008
[aaron@TC ~]$ which ruby
/Users/aaron/.rbenv/shims/ruby
[aaron@TC ~]$ ruby -v
ruby 2.4.0dev (2016-01-11 trunk 53499) [x86_64-darwin15]
[aaron@TC ~]$ rbenv shell 2.3.0
[aaron@TC ~]$ ruby -v
ruby 2.3.0p0 (2015-12-25 revision 53290) [x86_64-darwin15]
[aaron@TC ~]$ rbenv shell -
rbenv: OLD_RBENV_VERSION not set
[aaron@TC ~]$
```

I'd *expect* the above interaction to work and switch me back to `2.4.0`.  I gather that `RBENV_VERSION` is not set when I open the shell.  Is this by design?  Should we call this commit "good" and expect people to set `RBENV_VERSION` themselves?  FWIW, I usually do `rbenv shell ruby-trunk`, so I don't have a particular version and I don't mind setting the environment variable myself in the bashrc.

@mislav do you have any advice for me?

(Side note, I need to clean up the commit message so please don't merge this yet)

jasonkarns:
Maybe when old version hasn't yet been set, it should --unset (this would
get you back to whatever was in use prior to rbenv-shell).

tenderlove:
Looks like --unset [just removes the environment variable](https://github.com/rbenv/rbenv/blob/c38833179b368fc0ec044fdd9cdcba225e3fcdd2/libexec/rbenv-sh-shell#L39-L49).  I guess I'd need to understand how `libexec/rbenv` figures out the right version of Ruby in order to know what to put in `OLD_RBENV_VERSION` on the first run.  AFAICT, there isn't necessarily any state in the shell that will tell me what version of Ruby the user would run were they to try running it (but I could be totally wrong).

jeremy:
```
$ cd -
-bash: cd: OLDPWD not set
```

jasonkarns:
> Looks like --unset just removes the environment variable.

Right. So the way rbenv determines version is, in priority order:

1. env variable
2. local `.ruby-version` file in project directory, or any ancestor directory, up to root
3. global `version` file in `$(rbenv root)`

You can approximate this lookup yourself by checking `rbenv-shell`, then `rbenv-local` then `rbenv-global`. The first one that returns an actual version would be the ruby version used by rbenv.

So in your scenario:


1. new shell with `RBENV_VERSION` unset
2. ruby version is determined either by rbenv-local or rbenv-global (depending on if a local `.ruby-version` is found); let's say it's set to `2.4.0-dev` via rbenv-global
3. set shell: `rbenv shell 2.3.0`
    Now `RBENV_VERSION` is now 2.3.0, `OLD_RBENV_VERSION` is still unset
4. `rbenv shell -`
    Now we save off `RBENV_VERSION` to `OLD_RBENV_VERSION` (2.3.0);
    And `rbenv-shell --unset`, so rbenv would fall back to rbenv-global (2.4.0-dev) as in step 2
5. `rbenv version` -> 2.4.0-dev

As I work through this, I realize that we'll probably want `rbenv shell --unset` to save `RBENV_VERSION` off to `OLD_RBENV_VERSION` as well, in order to make repeated `rbenv shell -` work as intended. Continuing from above:

6. Another `rbenv shell -` would set `RBENV_VERSION` = `OLD_RBENV_VERSION` which is now '2.3.0'. And unset `OLD_RBENV_VERSION` (since `RBENV_VERSION` was previously unset). This puts us into the exact same state as 3 above and successfully makes repeated `rbenv shell -` invocations function in a loop.

tenderlove:
@jasonkarns derp. Thank you. That's much easier than I was thinking and doesn't require I know anything about how to look up versions (I think).  The only time we really need an error is if `OLD_RBENV_VERSION` and `RBENV_VERSION` are *both* unset and someone does `rbenv shell -`.

jasonkarns:
> The only time we really need an error is if OLD_RBENV_VERSION and RBENV_VERSION are both unset and someone does rbenv shell -.

yep, perfect!

jasonkarns:
While you're in there, I don't believe `OLD_RBENV_VERSION` needs to be exported, since it's not a variable that would need to be accessed in subshells.

sstephenson:
I love this!

libexec/rbenv-sh-shell
```diff
@@ -39,25 +39,50 @@ fi
...
     echo "unset RBENV_VERSION"
     ;;
   esac
   exit
 fi
 
-# Make sure the specified version is installed.
-if rbenv-prefix "$version" >/dev/null; then
+if [ "$version" = "-" ]; then
+  if [ -z "$OLD_RBENV_VERSION" ]; then
```
mislav: Because we don't intend to export OLD_RBENV_VERSION from the interactive shell, its value won't be accessible to this process. Instead, any operation (reading, writing) on OLD_RBENV_VERSION needs to be done from the code that is output from this process (and eval'd in the interactive shell).

libexec/rbenv-sh-shell
```diff
@@ -39,25 +39,50 @@ fi
 if [ "$version" = "--unset" ]; then
   case "$shell" in
   fish )
+    echo "set -e OLD_RBENV_VERSION"
```
mislav: Do we really want to erase the OLD_RBENV_VERSION? How about setting it to RBENV_VERSION that was before the `--unset`?

jasonkarns:
Just realized that the OLD_RBENV_VERSION variable name breaks the consistent variable namespace of `RBENV_*`. 

Will open a PR to change `OLD_RBENV_VERSION` to `RBENV_VERSION_OLD`, unless there's a reason not to?

jasonkarns:
Opened #958

PR #858, opened 2016-01-14
silence broken pipe errors from type/head

jasonkarns:
Occasionally (very rarely), I've encountered broken-pipe errors from rbenv-help. The underlying issue is that `head` can terminate and close its STDIN before the previous process in the pipeline is finished. If the upstream process doesn't properly handle SIGPIPE (or inherits SIG_IGN), then the upstream process can error out.

The scenario that I've encountered in rbenv-help is when `type -p gawk awk` is piped to `head`. Since the output of `type | head` is used as the command, occasionally the ouput is an error message about the broken pipe, rather than the expected `gawk`/`awk` command to use.

### DIFF
--- a/libexec/rbenv-help
+++ b/libexec/rbenv-help
@@ -42,7 +42,7 @@ extract_initial_comment_block() {
 }
 
 collect_documentation() {
-  $(type -p gawk awk | head -1) '
+  $(type -p gawk awk 2>/dev/null | head -1) '
     /^Summary:/ {
       summary = substr($0, 10)
       next


jasonkarns:
An example of this error has been captured in one of nodenv's [test runs on travis](https://travis-ci.org/OiNutter/nodenv/jobs/102413012). I don't think I've seen this issue locally but once.

mislav:
Thanks for letting me know, but I can't imagine how this error could have happened. Nobody has ever reported this, and our Travis CI didn't succumb to this. Is it possible that this is something that's only specific to nodenv test suite?

jasonkarns:
> Is it possible that this is something that's only specific to nodenv test suite?

I suppose anything is possible, but in this particular case I think it's unlikely. The help subcommand and associated test file is literally identical once the rbenv -> nodenv rename is accounted for. (Some areas of the codebase have actually forked, but the help subcommand is byte for byte the same.)

The one thing that did seem to start triggering this error more often (but still not consistently) was the conversion of our travis suite 'language' from C to NodeJS. Perhaps there's something in travis' node' environment which makes this error more common. Maybe the `type` built-in is slower in their container environment or something crazy.

PR #863, opened 2016-02-01
Updated Mac OSX brew install command

ErikAGriffin:
It seems rbenv now comes with ruby-build.  I have not investigated fully, but the previously shown command: `brew install rbenv ruby-build` caused issues on my machine.  After uninstalling both and simply running `brew install rbenv` everything worked fine.

### DIFF
--- a/README.md
+++ b/README.md
@@ -234,7 +234,7 @@ manager on Mac OS X:
 
 ~~~
 $ brew update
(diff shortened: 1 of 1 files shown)


mislav:
I don't see how the old command was a problem, but yes, ruby-build is a dependency of rbenv, so it gets auto-installed. Good catch.

PR #870, opened 2016-02-13
Update `rbenv init` instructions

davidcelis:
The README details `eval`ing `rbenv init -`, but for some shells (such as fish) there's a difference in what should be run. It turns out that `rbenv init` on its own will print correct instructions, so we should point users to running that command instead.

### DIFF
--- a/README.md
+++ b/README.md
@@ -180,13 +180,8 @@ easy to fork and contribute any changes back upstream.
 
     **Zsh note**: Modify your `~/.zshrc` file instead of `~/.bash_profile`.
 
-3. Add `rbenv init` to your shell to enable shims and autocompletion.
-
-    ~~~ sh
-    $ echo 'eval "$(rbenv init -)"' >> ~/.bash_profile
-    ~~~
-
-    _Same as in previous step, use `~/.bashrc` on Ubuntu, or `~/.zshrc` for Zsh._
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -180,13 +180,8 @@ easy to fork and contribute any changes back upstream.
...
     **Zsh note**: Modify your `~/.zshrc` file instead of `~/.bash_profile`.
 
-3. Add `rbenv init` to your shell to enable shims and autocompletion.
-
-    ~~~ sh
-    $ echo 'eval "$(rbenv init -)"' >> ~/.bash_profile
-    ~~~
-
-    _Same as in previous step, use `~/.bashrc` on Ubuntu, or `~/.zshrc` for Zsh._
+3. Run `rbenv init` for shell-specific instructions on how to initialize
```
mislav: Yeah, much better! Except, they've just cloned rbenv in `~/.rbenv` and added it to their PATH in the previous step, but most likely haven't restarted their shell. So, running just `rbenv` could result in "command not found". Maybe instruct them to run `~/.rbenv/bin/rbenv init` instead? It's a mouthful, but I don't have any other ideas.

davidcelis: Done!

mislav:
Thanks!

PR #873, opened 2016-02-17
Fish shell "." is deprecated in favor of "source"

dmerand:
Per [the fish documentation for "source"](http://fishshell.com/docs/current/commands.html#source):

```
(a single period) is an alias for the source command. The use of . is deprecated in favour of source, and . will be removed in a future version of fish."
```

### DIFF
--- a/libexec/rbenv-init
+++ b/libexec/rbenv-init
@@ -70,7 +70,7 @@ if [ -z "$print" ]; then
     echo
     case "$shell" in
     fish )
-      echo 'status --is-interactive; and . (rbenv init -|psub)'
+      echo 'status --is-interactive; and source (rbenv init -|psub)'
       ;;
     * )
       echo 'eval "$(rbenv init -)"'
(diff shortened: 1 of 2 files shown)


mislav:
Thanks for letting us know! Can you update the related test as well? See failure in Travis CI output.

libexec/rbenv-init
```diff
@@ -97,10 +97,7 @@ esac
 
 completion="${root}/completions/rbenv.${shell}"
 if [ -r "$completion" ]; then
-  case "$shell" in
-  fish ) echo ". '$completion'" ;;
-  *    ) echo "source '$completion'" ;;
-  esac
+  echo "source '$completion'"
```
mkllnk: Unfortunately  ash, dash and pdksh don't support `source`. At the moment I get this error on a Debian system:

      $ sh -c 'eval "$(rbenv init -)"'
      sh: 2: eval: source: not found

The `.` command is standardised POSIX:
http://pubs.opengroup.org/onlinepubs/9699919799/utilities/V3_chap02.html#dot

mkllnk: @mislav No, this pull request does not distinguish between shells at this code line. That's why I wouldn't merge it as is.

But you are right with my dash issue, I should file that.

PR #876, opened 2016-03-01
Remove references to .rbenv/bin/

iamvery:
:warning: I'm not an rbenv user, so this may be dumb :warning: 

Helping a student debug their Ruby environment this week, I ran into this issue trying to configure rbenv. The docs indicate that I need to run an init command, but no `bin/` directory exists in the rbenv installation. Running `rbenv init` instead seems to work and give us the further instructions we needed.

I'm happy to patch this up further if more is needed. Thanks!

### DIFF
--- a/README.md
+++ b/README.md
@@ -176,6 +176,8 @@ easy to fork and contribute any changes back upstream.
     $ echo 'export PATH="$HOME/.rbenv/bin:$PATH"' >> ~/.bash_profile
     ~~~
 
+    Then restart your terminal or reload your profile with `source ~/.bash_profile`.
+
     **Ubuntu Desktop note**: Modify your `~/.bashrc` instead of `~/.bash_profile`.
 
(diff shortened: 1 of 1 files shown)


iamvery:
I suspect the difference is install via Homebrew vs. installing from source. Homebrew adds the `rbenv` command to path for you. So from source, you'd still need to add `~/.rbenv/bin/` to path and reload your shell before running `rbenv init`.

mislav:
Thanks for the suggestions, but the part of the documentation that you amended applies only to method installation via git checkout to `~/.rbenv`, which is guaranteed to create a `~/.rbenv/bin` directory.

The Homebrew method is definitely a different and easier method of installation, but then these instructions don't apply. Instead, [these are the instructions to complete Homebrew installation](https://github.com/rbenv/rbenv#homebrew-on-mac-os-x), and they do suggest you run `rbenv init`.

iamvery:
Thanks for the taking your time to reply :heart:

PR #884, opened 2016-03-20
Adding check to make sure PATH variable is not populated multiple times.

oddraisin:
Add check to make sure that PATH isn't altered multiple times (e.g. once at login, a second time on tmux/screen shell, etc).


mislav:
Thanks for the suggestion, but we already had this feature and people didn't know how to deal with it. See explanation in e2173df4aa91c8d365ca1596fb857fcac9fdd787

PR #898, opened 2016-04-22
silence stdout when cding to working directory

mikelikesbikes:
I noticed an issue where the working directory was bring printed to STDOUT when running a script from outside the directory.

Here's more detail.

Given this project structure:

```
.
└── bin
    └── rubyscript
```

and the file `bin/rubyprogram` contains:

```ruby
#!/usr/bin/env ruby
puts "hi"
```

I see this in the console when I run the program from outside, then inside the `bin` directory.

```
$ bin/rubyprogram
/Users/mike/projects/ruby-script/bin
hi
$ cd ./bin
$ ./rubyprogram
hi
```

This PR fixes the issue.

### DIFF
--- a/libexec/rbenv
+++ b/libexec/rbenv
@@ -1,6 +1,6 @@
 #!/usr/bin/env bash
 set -e
-export -n CDPATH
+unset CDPATH
 
 if [ "$1" = "--debug" ]; then
   export RBENV_DEBUG=1


mislav:
Hi thanks for reporting. This. I'm very confused since `cd` usually doesn't output anything to stdout. Do you have something special configured on your system to trigger this?

mikelikesbikes:
@mislav to the best of my knowledge (and attempts to discover) I don't have anything set up to make cd echo it's target dir.

```
$ type cd
cd is a shell builtin
$ bash --version
GNU bash, version 4.3.33(1)-release (x86_64-apple-darwin14.1.0)
Copyright (C) 2013 Free Software Foundation, Inc.
License GPLv3+: GNU GPL version 3 or later <http://gnu.org/licenses/gpl.html>

This is free software; you are free to change and redistribute it.
There is NO WARRANTY, to the extent permitted by law.
```

When I run cd on it's own outside of this script, it doesn't echo either.

Any thoughts on how I can probe more into why cd might be behaving differently inside vs. outside of the rbenv script?

mikelikesbikes:
I think cd does this by itself...

from `~`
```
$ cd projects/ruby-script
/Users/mike/projects/ruby-script
$ cd ~/projects/ruby-script
(NO OUTPUT)
```

mikelikesbikes:
I have `CDPATH` set in my environment, and it appears, that when `cd` uses CDPATH (when target dir is relative, rather than absolute), it prints the full path of the directory it's changed to.

http://unix.stackexchange.com/questions/245453/how-to-not-show-path-after-cd-command-with-cdpath-set

I looked more at that in the script... here's the first 3 lines:

```
#!/usr/bin/env bash
set -e
export -n CDPATH
```

I don't think the `export -n CDPATH` is working as expected. Do you agree that the original `export -n CDPATH` isn't having the desired effect (ie. `cd` is still behaving as if `CDPATH` was set)?

Both of the following changes fix the issue I was seeing (without the need for the commit already in this PR):

```
#!/usr/bin/env -uCDPATH bash
set -e
```
or
```
#!/usr/bin/env bash
set -e
export CDPATH=""
```

Any preference on which of the above 2 you like better (prefer `export CDPATH=""`)?

mikelikesbikes:
Another option, my actual favorite:

```
#!/usr/bin/env bash
set -e
unset CDPATH
```

mikelikesbikes:
Curiously,

`rbenv/bin/rbenv` already had this... but `rbenv/libexec/rbenv` did not...

(sorry for all the chatter on this)

mislav:
No need to apologize for the chatter; I'm glad you tracked this down.

If you change `export -n CDPATH` to `unset CDPATH`, does it fix the issue?

mikelikesbikes:
Yes. `unset CDPATH` also fixes the original issue I noticed. I updated my branch (removing the old commit and replacing it with [this fix](https://github.com/rbenv/rbenv/pull/898/commits/d246e20764dbbeea4fa900d6bb2652eadcb996e6) instead)

mislav:
Thanks!

PR #901, opened 2016-05-10
Prevent duplicate ~/.rbenv/shims in $PATH

ymattw:
Note fish does not like ${var} notation, and here I use the POSIX
substring parameter expansion to test if $RBENV_ROOT is a substring of
$PATH.

It does not work for csh but that's an existing problem and should be
fixed separately.

### DIFF
--- a/libexec/rbenv-init
+++ b/libexec/rbenv-init
@@ -86,11 +86,11 @@ mkdir -p "${RBENV_ROOT}/"{shims,versions}
 
 case "$shell" in
 fish )
-  echo "setenv PATH '${RBENV_ROOT}/shims' \$PATH"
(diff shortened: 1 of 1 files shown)


mislav:
We already had this, but had to revert since it conflicted with too many people's environments (most often tmux) https://github.com/rbenv/rbenv/commit/e2173df4aa91c8d365ca1596fb857fcac9fdd787

ymattw:
We still have chance to make it better, how about remove the target component `${RBENV_ROOT}/shims` from PATH before prepending to it when `rbenv init -` is called?

mislav:
No I have decided that a "smart" approach doesn't really buy us anything, and could cause unpredictable results in people's weird environments. I would just count on people cleaning up PATH themselves when it gets messy for them. I don't think we should be responsible for it, and—to my knowledge—we don't have anything to gain by removing the duplicate ourselves.

ymattw:
Well said, I buy that.

PR #914, opened 2016-07-01
Help message should not be written to stdout for `sh-` commands

yyuu:
`rbenv shell --help` is broken since it tries to write help messages on stdout. stdout for `sh-` commands should be shell commands though.

```
% rbenv shell --help
(eval):2: parse error near `\n'
```

As a workaround, this patch will redirect help messages to stderr instead of stdout.

### DIFF
--- a/libexec/rbenv
+++ b/libexec/rbenv
@@ -109,7 +109,11 @@ case "$command" in
 
   shift 1
   if [ "$1" = --help ]; then
-    exec rbenv-help "$command"
+    if [[ "$command" == "sh-"* ]]; then
+      exec rbenv-help "$command" 1>&2
+    else
+      exec rbenv-help "$command"
+    fi
   else
     exec "$command_path" "$@"
   fi
--- a/libexec/rbenv
+++ b/libexec/rbenv
@@ -110,7 +110,7 @@ case "$command" in
   shift 1
   if [ "$1" = --help ]; then
     if [[ "$command" == "sh-"* ]]; then
-      exec rbenv-help "$command" 1>&2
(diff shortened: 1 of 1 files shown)


mislav:
Good catch! This could be one workaround, but I'm not happy with the fact that going forward, `rbenv shell --help` will be mysteriously output to stderr, while all others go to stdout. An explicit call for `--help` should always go to stdout, so that a program could capture help text (in theory) and pass it on somewhere else; such as to a pager (e.g. `less`).

Maybe the workaround belongs instead in the `rbenv()` shell wrapper function. But that might make the function more complex, which is tricky to pull off since the function needs to work for a variety of shells (including fish, which has a totally different syntax).

yyuu:
You're right. I didn't want to make `rbenv()` more complex.

I reconsider about the implementation and prepared 26ac59fd1da0a28695697a03bd575810bc962d90. Now it should work fine with stdout. Please review it.

libexec/rbenv
```diff
@@ -109,7 +109,11 @@ case "$command" in
 
   shift 1
   if [ "$1" = --help ]; then
-    exec rbenv-help "$command"
+    if [[ "$command" == "sh-"* ]]; then
+      echo "rbenv help \"$command\""
```
mislav: That's pretty sneaky, but it solves the problem!

I guess this might interfere is somone goes straight to something like `rbenv sh-shell --help`. Instead of help text, the echo statement would be output. However, since `sh-` prefix is considered an implementation detail and shouldn't be referenced directly from the UI, I don't think we should be worried about this. 👍

PR #919, opened 2016-08-01
Add a workaround system executable in `/bin`

yyuu:
It seems that `rbenv-prefix` doesn't work well if the `system` executable is installed in `/` (`/bin`).

### DIFF
--- a/test/prefix.bats
+++ b/test/prefix.bats
@@ -25,12 +25,13 @@ load test_helper
 }
 
 @test "prefix for system in /" {
-  mkdir -p "${RBENV_TEST_DIR}/bin"
-  touch "${RBENV_TEST_DIR}/bin/rbenv-which"
-  echo "echo /bin/ruby" >"${RBENV_TEST_DIR}/bin/rbenv-which"
-  chmod +x "${RBENV_TEST_DIR}/bin/rbenv-which"
+  mkdir -p "${BATS_TEST_DIRNAME}/libexec"
+  touch "${BATS_TEST_DIRNAME}/libexec/rbenv-which"
+  echo "echo /bin/ruby" >"${BATS_TEST_DIRNAME}/libexec/rbenv-which"
+  chmod +x "${BATS_TEST_DIRNAME}/libexec/rbenv-which"
   RBENV_VERSION="system" run rbenv-prefix
   assert_success "/"
+  rm -f "${BATS_TEST_DIRNAME}/libexec/rbenv-which"
 }
 
 @test "prefix for invalid system" {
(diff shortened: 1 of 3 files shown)


test/test_helper.bash
```diff
@@ -20,10 +20,10 @@ if [ -z "$RBENV_TEST_DIR" ]; then
   export RBENV_HOOK_PATH="${RBENV_ROOT}/rbenv.d"
 
   PATH=/usr/bin:/bin:/usr/sbin:/sbin:/usr/local/bin
-  PATH="${RBENV_TEST_DIR}/bin:$PATH"
   PATH="${BATS_TEST_DIRNAME}/../libexec:$PATH"
   PATH="${BATS_TEST_DIRNAME}/libexec:$PATH"
   PATH="${RBENV_ROOT}/shims:$PATH"
+  PATH="${RBENV_TEST_DIR}/bin:$PATH"
```
mislav: Instead of reordering the PATH here, which is risky, could you instead write a temporary `rbenv-which` stub to `${BATS_TEST_DIRNAME}/libexec`? Just make sure the test cleans up afterwards.

yyuu: Ok, I will try to do so.

test/prefix.bats
```diff
@@ -24,6 +24,16 @@ load test_helper
   assert_success "$RBENV_TEST_DIR"
 }
 
+@test "prefix for system in /" {
+  mkdir -p "${BATS_TEST_DIRNAME}/libexec"
+  touch "${BATS_TEST_DIRNAME}/libexec/rbenv-which"
```
mislav: You don't need this `touch` line.

test/prefix.bats
```diff
@@ -24,6 +24,16 @@ load test_helper
...
+  echo "echo /bin/ruby" >"${BATS_TEST_DIRNAME}/libexec/rbenv-which"
```
mislav: I didn't know executables could run without a shebang line! 😲

test/prefix.bats
yyuu: Ah, I'll remove it.

test/prefix.bats
yyuu: I think an executable without shebang will be called with `/bin/sh` by execve(2), but could not find any specification about it.... Just in case I'll add shebang line for it since the feature may be platform dependent.

http://stackoverflow.com/questions/12296308/shell-script-working-fine-without-shebang-line-why

yyuu:
Done!

mislav:
Thanks!!

Issue #301, opened 2012-12-31
`rbenv update`


mislav:
Should we have an [`rbenv update`](https://github.com/rkh/rbenv-update/blob/master/bin/rbenv-update) command? /cc @rkh

It would work only if `RBENV_ROOT` is where rbenv is installed, and if it's a git repo.

I think it might be useful, considering the slow release cycle, and especially now that we have `rbenv --version` capable of displaying edge versions from git.

wadkar:
@mislav , if it's okay - I recommend that we "get inspired" from [`pyenv`'s update plugin](https://github.com/pyenv/pyenv-update). Pyenv folks are nice and they consider `rbenv{,-build}` as upstream. So it will be a nice fit.

I have also read through the [`pyenv-update` bash script](https://github.com/pyenv/pyenv-update/blob/master/bin/pyenv-update) and I think it's pretty solid in terms of quoting things, and making sure things don't blow up. We just need to change few variables from `PYENV_ROOT` to `RBENV_ROOT`.

Do you want it to be part of `rbenv` core or as a separate plugin? In either case, I will be happy to submit a PR/repo.

cornfeedhobo:
I've started to use [`basher`](https://github.com/basherpm/basher/) to manage these tools.

Otherwise, my +1 is for what @wadkar said; a rbenv-update plugin.

olivierlacan:
Every time I run `rbenv instlal <new_version>` and I get this message I think: 
> We're so, so close to a nice experience for Ruby upgrades, it's really a shame rbenv doesn't just run this update in the background

```
ruby-build: definition not found: 3.0.1

See all available versions with `rbenv install --list'.

If the version you need is missing, try upgrading ruby-build:

  git -C /Users/olivierlacan/.rbenv/plugins/ruby-build pull
```

Having to run `git pull` is not the end of the world, but considering a tool like Homebrew routinely does it behind its own commands, it would really make a ton of sense to just run it (and be clear that rbenv is running a `git pull`).

mislav:
@olivierlacan I do believe that this should be easier for the user, and what you're suggesting sounds completely reasonable.

However, there's something that feels slightly risky about software that updates itself by itself without any intervention from the user—especially when the software has never done so before. I'm not sure if we should alienate our users' trust by automatically updating ruby-build on every `rbenv install`.

[...]

PR #937, opened 2016-10-07
Adding Rbenv update command

bekicot:
Adding `rbenv update` command for easy updating via git.

My first shell code pull request, please let me know any feedback :)


bekicot:
@mislav didn't see #301 before. I couldn't find easy way to update, so i created this. Is the feature is already created somewhere. :)

jasonkarns:
There's a plugin: https://github.com/rkh/rbenv-update

hsbt:
We don't need to integrate `rbenv update` because we relied on plugin system.

PR #940, opened 2016-10-19
friendlier error message with suggestion

Fenntasy:
Nothing fancy but just give out the command for the forgetful

### DIFF
--- a/libexec/rbenv-version-name
+++ b/libexec/rbenv-version-name
@@ -32,6 +32,6 @@ elif version_exists "${RBENV_VERSION#ruby-}"; then
 else
   echo "rbenv: version \`$RBENV_VERSION' is not installed (set by $(rbenv-version-origin))" >&2
   echo "install it with:" >&2
-  echo -e  "\t rbenv install $RBENV_VERSION" >&2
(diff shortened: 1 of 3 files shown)


mislav:
Hi, thanks for the contribution! Your improvement is perfectly valid in the case when there is ruby-build installed in addition to rbenv and available as `rbenv install` command. However, ruby-build component has always been optional, and we can't count on its presence. Some people might choose to have a different tool to install Rubies, and some people may simply have scripts that automate compiling Rubies into the `versions/` directory. Therefore, we can't suggest that people use `rbenv install` in an error message, since we don't know whether they would be able to run that command.

Fenntasy:
Would it be possible to detect if ruby-build is installed?

mislav:
Perhaps, at the time of generating this error message, you could check `command -v rbenv-install` in PATH. This would yield the path to the executable, or fail when none is found. However, I'd prefer to keep the error message in its original format. It shouldn't be hard to figure out how to install Rubies! We document it in the README, after all.

Fenntasy:
Ok, thanks for your time :)

PR #955, opened 2016-11-29
Support Msys2 on Windows

nobitagamer:
This is workaround for rbenv to work on MSys/Msys2 on Windows.

### DIFF
--- a/bin/rbenv
+++ b/bin/rbenv
@@ -1,1 +1,4 @@
-../libexec/rbenv
+pushd `dirname $0` > /dev/null
+SCRIPTPATH=`pwd`
+popd > /dev/null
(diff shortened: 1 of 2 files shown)


mislav:
We never had a goal to support Windows. I'm reluctant to go down this road.

1. Are you saying that once you get rid of the symlink, every other feature of rbenv works?
2. Why did you commit the `.patch` file?

hsbt:
And It's hard to test and maintain continuously for us. Because Msys2 on Windows environment is not FREE box.

mislav:
Sorry, we won't support Windows for now because we don't have a system in which we can continuosly test our compatibility with Windows in the long run. Also, it's very likely that more functionality than just this will be broken on Windows.

PR #958, opened 2016-12-19
Conforms OLD_RBENV_VERSION to RBENV_* convention

jasonkarns:
Keeping rbenv-controlled variables to RBENV_* "namespace" helps with
discoverability (and tools like rbenv-env) but also consistency and a
very minor degree of safety/isolation from env impact.


mislav:
Makes sense. Thanks!

PR #959, opened 2017-01-02
Fixed `rbenv init -` output to work w/ no args and bash's `set -u`.

zenspider:
If you use `set -u` in your bash setup (error on uninitialized vars) then `rbenv` will fail w/ no args because `$1` is used and unset. This fixes that.

I have not attempted to root out any others yet.

### DIFF
--- a/libexec/rbenv-init
+++ b/libexec/rbenv-init
@@ -138,7 +138,7 @@ esac
 if [ "$shell" != "fish" ]; then
 IFS="|"
 cat <<EOS
-  command="\$1"
+  command="\${1:-}"
   if [ "\$#" -gt 0 ]; then
     shift
   fi


mislav:
Thanks! Would this syntax be evaluated the same in other shells such as csh, ksh? The `rbenv()` function needs to stay as simple as possible because, unlike the rest of the codebase, it gets evaluated in many different shells.

zenspider:
I can't speak for other shells... 99% sure it'd be fine in zsh. No clue on csh. Might want to switch to a `-z` check? Dunno what the best practice is.

masudaK:
I escaped this to write below anyway.

```
set +u
rbenv shell VERSION
set -u
```

zenspider:
Yup... That is what I'm stuck with too

zenspider:
I just rebased this against current master... I'd really like to not run a private version..

PR #966, opened 2017-02-03
Fix typo in readme

fbnlsr:


kylef:
I do not think this is a typo, `rbenv` is the name of this project and the subcommand `install` does not come out of the box with `rbenv`.

mislav:
Thanks for the explanation, @kylef.

PR #968, opened 2017-02-10
Fix fish subcommand completion

maxnordlund:
This allows subcommand style plugins to properly autocomplete.
Existing commands are not affected. 

Example, say you have support for `rbenv foo bar --flag`, then
this allows the last `--flag` argument to be properly completed.

See also https://github.com/yyuu/pyenv/pull/831

### DIFF
--- a/completions/rbenv.fish
+++ b/completions/rbenv.fish
@@ -18,5 +18,6 @@ end
 
 complete -f -c rbenv -n '__fish_rbenv_needs_command' -a '(rbenv commands)'
 for cmd in (rbenv commands)
-  complete -f -c rbenv -n "__fish_rbenv_using_command $cmd" -a "(rbenv completions $cmd)"
+  complete -f -c pyenv -n "__fish_pyenv_using_command $cmd" -a \
+    "(rbenv completions (commandline -opc)[2..-1])"
 end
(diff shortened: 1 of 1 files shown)


completions/rbenv.fish
```diff
@@ -18,5 +18,6 @@ end
 
 complete -f -c rbenv -n '__fish_rbenv_needs_command' -a '(rbenv commands)'
 for cmd in (rbenv commands)
-  complete -f -c rbenv -n "__fish_rbenv_using_command $cmd" -a "(rbenv completions $cmd)"
+  complete -f -c pyenv -n "__fish_pyenv_using_command $cmd" -a \
```
znz: `pyenv` → `rbenv` ? (2 places)

maxnordlund: Whoops, I'll fix it right away.

completions/rbenv.fish
```diff
@@ -18,5 +18,6 @@ end
 
 complete -f -c rbenv -n '__fish_rbenv_needs_command' -a '(rbenv commands)'
 for cmd in (rbenv commands)
-  complete -f -c rbenv -n "__fish_rbenv_using_command $cmd" -a "(rbenv completions $cmd)"
+  complete -f -c rbenv -n "__fish_rbenv_using_command $cmd" -a \
+    "(rbenv completions (commandline -opc)[2..-1])"
```
mislav: What does the expression `(commandline -opc)[2..-1]` signify?

maxnordlund: It's the part after `pyenv` that the user has currently typed in, like `pyenv <stuff> <here> <etc>`. It's safe to do this slice, because `__fish_rbenv_using_command` already checks it's at least 2 long.

PR #972, opened 2017-02-24
Use long option for make -C/--directory

ElijahLynn:
Self documenting and easier to read and understand what is happening with long options.


mislav:
Thank you and even though long options are better to see in documentation in most cases, in this case I would opt for having the example short. I also think it's evident from earlier on the same line that `src` is a directory, and therefore the flag is self-explanatory even for someone who is not familiar with `make` flags.

PR #981, opened 2017-03-23
Better error message for `rbenv shell`

mhinz:
Shell integration is not enabled by default. This means that, from all the commands from `rbenv commands`, only "shell" won't work right away.

Replace "no such command" with a more descriptive message that points to `rbenv init` instead.

### DIFF
--- a/libexec/rbenv
+++ b/libexec/rbenv
@@ -106,7 +106,13 @@ case "$command" in
   ;;
 * )
   command_path="$(command -v "rbenv-$command" || true)"
-  [ -n "$command_path" ] || abort "no such command \`$command'"
+  if [ -z "$command_path" ]; then
+    if [ "$command" == "shell" ]; then
+      abort "shell integration not enabled. Run \`rbenv init' for instructions."
+    else
+      abort "no such command \`$command'"
+    fi
+  fi
 
   shift 1
   if [ "$1" = --help ]; then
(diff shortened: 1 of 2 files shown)


libexec/rbenv
```diff
@@ -105,7 +105,13 @@ case "$command" in
   ;;
 * )
   command_path="$(command -v "rbenv-$command" || true)"
-  [ -n "$command_path" ] || abort "no such command \`$command'"
+  if [ -z "$command_path" ]; then
+    if [ "$command" == "shell" ]; then
```
mislav: Could we have a test that covers this case?

libexec/rbenv
```diff
@@ -105,7 +105,13 @@ case "$command" in
...
+      abort "shell integration not enabled: rbenv init"
```
mislav: It's not clear that `rbenv init` is something that should be run. Maybe make the message more clear with something like:

```sh
abort "shell integration not enabled. Run \`rbenv init' for instructions."
```

mhinz:
Okay, I changed the wording of the error message, added tests, and rebased.

ADTC:
I came across this because apparently I'm on `rbenv` 1.1.1 rather than 1.1.2. Is there any reason why `apt install rbenv` installed the outdated version for me?

PR #982, opened 2017-03-27
partly resolve #759

### LINKED ISSUE #759: rbenv: cannot rehash: /usr/local/var/rbenv/shims/.rbenv-shim exists
Hi! When trying to `$ rbenv rehash` getting:

```
rbenv: cannot rehash: /usr/local/var/rbenv/shims/.rbenv-shim exists
```

What does that mean?

MartinNowak:
- some versions of bash (e.g. 4.3.11) complain about clobbering
  /dev/null, use bash's explicit >| operator to ignore noclober
- not sure if the original #759 report has the same cause
- https://github.com/rbenv/rbenv/issues/759#issuecomment-289326891


mislav:
Good catch; thank you!

Issue #866, opened 2016-02-03
CDPATH can result in incorrect resolution of RBENV_DIR


tpope:
[This `cd`](https://github.com/rbenv/rbenv/blob/d6ab394efde6871d0af8945d703cde1ef72d317a/libexec/rbenv#L63) can end up in the wrong place (plus print that wrong place to stdout) if the directory in question is relative and in `CDPATH` (which is unexported but not unset at the top of the file). Obvious workaround is to qualify it, perhaps with one of the following (or something more elegant);
- `RBENV_DIR=$(abs_dirname "$RBENV_DIR/.")`
- `[[ $RBENV_DIR == /* ]] || RBENV_DIR=./$RBENV_DIR`

I have further thoughts on `export -n CDPATH` which do not belong in this issue, but suffice to say I will argue that `unset CDPATH` is another step in the wrong and direction so am advocating against it as a solution here.

mislav:
Thanks for the heads-up. Here is some background that argued against CDPATH: #316

What would be your proposed solution? To leave CDPATH alone and just be careful not to `cd` to any relative path not prefixed with `./`?

tpope:
I'm afraid I don't have anything more elegant than that.

My problem with the advice "just set `CDPATH` in `.bashrc` and don't export it" is twofold:

1. More than just the shell respects `CDPATH` (e.g. Vim)
2. Sometimes it is useful to set a value for a non-global context (e.g. a tmux session)

I was considering proposing a compromise along the lines of "leave it exported for the exec subcommand only", since I have not encountered a case where the other built-ins spawn a process where it would matter.

blueyed:
@mislav 
I cannot see from #316 why it should not be unset (instead of only not being exported).
See also https://github.com/rbenv/rbenv/commit/e3f72ebae20768079ca4b4425a364900f3f16fc6, where you've mentioned this fact yourself.

PR #985, opened 2017-04-05
Don't unset CDPATH

tpope:
I strongly feel that unsetting `CDPATH` was an overreaction to an easily addressed edge case, and have practical reasons for wanting to get it restored. I know this is a contentious issue, so here's a comprehensive analysis of the 5 uses (9 invocations) of `cd` in `libexec`.

* [`abs_dirname` in `rbenv`](https://github.com/rbenv/rbenv/blob/4f8925abe7e4b373156bba9cc5310e8e0d04a28a/libexec/rbenv#L38--50), which is only ever [called on `$0`](https://github.com/rbenv/rbenv/blob/4f8925abe7e4b373156bba9cc5310e8e0d04a28a/libexec/rbenv#L72), which is absolute unless you do something like `cd ~/.rbenv; bin/rbenv` (can be fixed if it's an actual problem).
* [`$RBENV_DIR` in `rbenv`](https://github.com/sstephenson/rbenv/blob/4f8925abe7e4b373156bba9cc5310e8e0d04a28a/bin/rbenv#L63), the culprit from #866 which provoked the unset. Addressed in this PR.
* [`${BASH_SOURCE%/*}` in `rbenv---version`](https://github.com/sstephenson/rbenv/blob/4f8925abe7e4b373156bba9cc5310e8e0d04a28a/libexec/rbenv---version#L18), which will [be absolute when called by `rbenv`](https://github.com/sstephenson/rbenv/blob/4f8925abe7e4b373156bba9cc5310e8e0d04a28a/bin/rbenv#L107-L119).
* [`realpath` in `rbenv-hooks`](https://github.com/sstephenson/rbenv/blob/4f8925abe7e4b373156bba9cc5310e8e0d04a28a/libexec/rbenv-hooks#L39-L52), which is only ever called on [an absolute hook directory](https://github.com/sstephenson/rbenv/blob/4f8925abe7e4b373156bba9cc5310e8e0d04a28a/libexec/rbenv-hooks#L55-L63).
* [`realpath` in `rbenv-versions`](https://github.com/sstephenson/rbenv/blob/4f8925abe7e4b373156bba9cc5310e8e0d04a28a/libexec/rbenv-versions#L46-L59), which is only ever called on [an absolute version directory](https://github.com/sstephenson/rbenv/blob/4f8925abe7e4b373156bba9cc5310e8e0d04a28a/libexec/rbenv-versions#L62-L64).

That just leaves plugins. In my humble opinion, plugins can `unset CDPATH` themselves if need be, but if that's a sticking point, let me know, and maybe we can compromise on leaving `CDPATH` in place for `rbenv-exec`, or something.

### DIFF
--- a/libexec/rbenv
+++ b/libexec/rbenv
@@ -60,6 +60,7 @@ export RBENV_ROOT
 if [ -z "${RBENV_DIR}" ]; then
   RBENV_DIR="$PWD"
 else
+  [[ $RBENV_DIR == /* ]] || RBENV_DIR=$PWD/$RBENV_DIR
   cd "$RBENV_DIR" 2>/dev/null || abort "cannot change working directory to \`$RBENV_DIR'"
   RBENV_DIR="$PWD"
   cd "$OLDPWD"
--- a/libexec/rbenv
+++ b/libexec/rbenv
@@ -1,6 +1,5 @@
 #!/usr/bin/env bash
 set -e
-unset CDPATH
 
 if [ "$1" = "--debug" ]; then
   export RBENV_DEBUG=1


mislav:
Thanks for the research!

blueyed:
Another option, and maybe a good idea in general, would be using `set -p` for privileged mode (see `man bash`).

PR #986, opened 2017-04-07
Prefer 'set' over 'setenv' for fish shell

jeffkowalski:
The setenv function in fish shell has changed dramatically in
https://github.com/fish-shell/fish-shell/commit/75600b6b538f242a6b340e51965f25cfd4473c4a
It now conforms to the csh version, which takes at most two arguments.
In this init script, the form
    setenv PATH prepend_something $PATH
had been used, which had too many arguments.
Since setenv isn't a native command in fish, a suitable replacement is
to use the "set -gx" command, which can consume multiple arguments.

### DIFF
--- a/libexec/rbenv-init
+++ b/libexec/rbenv-init
@@ -86,8 +86,8 @@ mkdir -p "${RBENV_ROOT}/"{shims,versions}
 
 case "$shell" in
 fish )
-  echo "setenv PATH '${RBENV_ROOT}/shims' \$PATH"
-  echo "setenv RBENV_SHELL $shell"
+  echo 'set -gx PATH '${RBENV_ROOT}'/shims $PATH'
+  echo "set -gx RBENV_SHELL $shell"
 ;;
 * )
   echo 'export PATH="'${RBENV_ROOT}'/shims:${PATH}"'
--- a/libexec/rbenv-init
+++ b/libexec/rbenv-init
@@ -86,7 +86,7 @@ mkdir -p "${RBENV_ROOT}/"{shims,versions}
 
 case "$shell" in
 fish )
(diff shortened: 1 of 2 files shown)


asymmetric:
Because of this, currently rbenv is broken on fish 2.6b1, which is the latest beta.

libexec/rbenv-init
```diff
@@ -86,8 +86,8 @@ mkdir -p "${RBENV_ROOT}/"{shims,versions}
 
 case "$shell" in
 fish )
-  echo "setenv PATH '${RBENV_ROOT}/shims' \$PATH"
-  echo "setenv RBENV_SHELL $shell"
+  echo 'set -gx PATH '${RBENV_ROOT}'/shims $PATH'
```
asymmetric: Why the change in quotes?

jeffkowalski: The change reflects the quoting style of the default ```*)``` case just below at line 93.

asymmetric: But the case below is bash. I would rather keep the quote change in another PR, if need be. Have you tested this btw?

The CI tests seem to fail.

mislav: @jeffkowalski Whatever quoting style you pick in the end, please ensure that the accompanying test is also updated and that CI passes again. Thanks!

jeffkowalski:
New commit a81da8d864c46c82b8d4ddc8da778f2cdf4274dd reverts the quote changes and fixes the init.bat.  Travis CI now passes.

mislav:
Thank you!

rahearn:
Is there a plan to release a new Homebrew version with this change? Fish 2.6.0 is released now, so the published version of rbenv no longer works. Thanks

PR #999, opened 2017-05-16
Make homebrew installation instructions more clear

jacob-on-github:
I work on a team that has followed the Homebrew installations. More than once we've missed the `rbenv init` instruction and it has caused headaches down the road. This formatting makes it harder to miss.


mislav:
Looks good; thank you!

PR #1002, opened 2017-06-05
test/which.bats: do not export PATH

blueyed:
Otherwise `sed` might not be found later in bats/libexec/bats-exec-test,
in case `/usr/bin` gets removed.


blueyed:
Updated to not use `env` anymore - where `run` was not found then.

mislav:
Each test is run in isolation (I think) as a separate process. But the change looks good anyway.

PR #1009, opened 2017-06-13
Clarify Homebrew/git install guidance

jeremy:
* Move Homebrew to the top, common case for new devs
* Note that `brew install rbenv` includes ruby-build by default
* Be clear that `rbenv init` instructions are needed for shell setup
* Be explicit about starting a new shell to pick up PATH/init changes
* Use rbenv-doctor to conclusively demonstrate correct setup
* Separate upgrade instructions for Homebrew vs Git installs
    
Related: https://github.com/rbenv/ruby-build/pull/1106

### DIFF
--- a/README.md
+++ b/README.md
@@ -39,9 +39,10 @@ RVM?**](https://github.com/rbenv/rbenv/wiki/Why-rbenv%3F)
   * [Choosing the Ruby Version](#choosing-the-ruby-version)
   * [Locating the Ruby Installation](#locating-the-ruby-installation)
 * [Installation](#installation)
+  * [Homebrew on macOS](#homebrew-on-mac-os)
+    * [Upgrading with Homebrew](#upgrading-with-homebrew)
   * [Basic GitHub Checkout](#basic-github-checkout)
-    * [Upgrading](#upgrading)
-  * [Homebrew on Mac OS X](#homebrew-on-mac-os-x)
+    * [Upgrading with Git](#upgrading-with-git)
   * [How rbenv hooks into your shell](#how-rbenv-hooks-into-your-shell)
   * [Installing Ruby versions](#installing-ruby-versions)
     * [Installing Ruby gems](#installing-ruby-gems)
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -148,15 +149,61 @@ Version names to rbenv are simply the names of the directories in
...
+   running `ruby` "see" the Ruby version that you choose with rbenv.
+
+3. Close your Terminal window and open a new one so your changes take
+   effect.
+
+4. Verify that rbenv is properly set up using this
+   [rbenv-doctor](https://github.com/rbenv/rbenv-installer/blob/master/bin/rbenv-doctor) script:
+
+    ~~~ sh
+    $ curl -fsSL https://raw.githubusercontent.com/rbenv/rbenv-installer/master/bin/rbenv-doctor | bash
```
mislav: The `rbenv-doctor` step takes up a lot of vertical space and is repeated twice. I'd be in favor of instead mentioning it only once by linking to the `rbenv-installer` repo where there are instructions how to run the rbenv doctor.

jeremy: I did that at first, but I don't think we benefit from DRY here.

This is super-concrete, exactly what users need to know and do. Repeating it is appropriate because users are paying attention to the task at hand, not perusing the complete document.

mislav: While we're following redirects with `-L`, let's use the canonical URL for raw files, since `raw.githubusercontent.com` is an implementation detail: `https://github.com/rbenv/rbenv-installer/raw/master/bin/rbenv-doctor`

README.md
```diff
@@ -148,15 +149,61 @@ Version names to rbenv are simply the names of the directories in
...
+    Checking for `rbenv' in PATH: /usr/local/bin/rbenv
+    Checking for rbenv shims in PATH: OK
+    Checking `rbenv install' support: /usr/local/bin/rbenv-install (ruby-build 20170523)
+    Counting installed Ruby versions: 14 versions
```
mislav: A fresh install of rbenv will not have any rbenv versions installed. Do we want to match here what the user will likely see?

jeremy: Waffled on this, too. Agreed, will start with none.

PR #1011, opened 2017-06-19
Properly escapes commas in c flags

kenaniah:
Solves issue with sed replacement for `@CFLAGS@` on GNU Linux platforms during the configure stage by escaping commas.

### DIFF
--- a/src/configure
+++ b/src/configure
@@ -35,12 +35,12 @@ eval "$("$src_dir"/shobj-conf -C "$CC" -o "$host_os")"
 
 sed "
   s,@CC@,${CC},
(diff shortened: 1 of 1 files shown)


mislav:
Hi, thanks for this patch! Can you elaborate a bit more about when do commas appear in CFLAGS and what happens if they're not escaped? Thank you!

kenaniah:
Commas appear in CFLAGS when configuring on [Solus OS](https://solus-project.com/download/) (a GNU Linux variant). The result is that `sed` throws a syntax error which causes `make` to fail.

PR #1013, opened 2017-06-23
Fix some issues reported by shellcheck

blueyed:


blueyed:
Should be squashed when merging.

libexec/rbenv-help
```diff
@@ -112,9 +113,9 @@ print_help() {
   local command="$1"
   local summary usage help
   eval "$(documentation_for "$command")"
-  [ -n "$help" ] || help="$summary"
+  [[ -n "$help" ]] || help="$summary"
```
mislav: The double `[[` here are not necessary.

libexec/rbenv-help
```diff
@@ -112,9 +113,9 @@ print_help() {
...
 
-  if [ -n "$usage" -o -n "$summary" ]; then
+  if [[ -n "$usage" ]] || [[ -n "$summary" ]]; then
```
mislav: Ditto `[[`

libexec/rbenv-help
blueyed: Not necessary, but better in fact (`[` is a command, `[[` an instruction), no?

mislav: In bash, `[` is a builtin (see `help [`), therefore there is no performance saving.

We only use `[[` for pattern matching.

blueyed: I see, this is different in Zsh then (AFAIK).

Amended / squashed / rebased.

PR #1024, opened 2017-08-07
Fixed support for commas for all environment variables in configure script

andyone:
Hi! Currently in configure script comma used as a separator in sed command. Sed allows using any character as a separator. I recommend using number sign (`#`) as a separator. In this case, we don't need to escape the commas in environment variables.


mislav:
Thank you! Now we have to hope that none of the values will include `#` 😝 

Followup to https://github.com/rbenv/rbenv/pull/1011

Issue #1000, opened 2017-05-16
GitHub Pages for Documentation


jacobherrington:
It looks like the gh-pages branch was abandoned. Is this something that should be created going forward? As a user, it feels like the documentation could be presented better and building a page for documentation might be the best way to do that

jacob-on-github:
You can see the abandoned site at rbenv.org.

jacob-on-github:
I will work on this unless I hear otherwise.

mislav:
You're welcome to suggest some documentation. How did you plan to document things?

We would probably want to have said documentation in a separate repo named `rbenv.github.io`. But you're welcome to send a PR for the gh-pages branch in this repo first.

jacob-on-github:
@mislav I plan on porting the readme into an easy to follow "getting started," then look at expanding it to cover rbenv's features.

I'm thinking something Jekyll based, but I'll be sure to get your feedback on the PR.

Grizel2202:
Thanks

PR #1027, opened 2017-08-21
built jekyll site to house documentation

jacobherrington:
I built a basic Jekyll site to house documentation. The copy might need to be updated/changed based on what needs to be available for users and contributors to become familiar with rbenv.

refs #1000


mislav:
Thank you for your suggestion. What is the benefit of the documentation site over the current README?

jacobherrington:
@mislav As rbenv continues to grow in may be beneficial to have something representing the tool online outside of GitHub - especially if that site can help new developers get started with rbenv. I personally had trouble following along with the readme and would have preferred some kind of intro tutorial or body of work outside of the readme. Of course, that's just my opinion.

PR #1038, opened 2017-09-18
Improve Fish compatiblity with subshells

kennethreitz:
This increases compatibility with subshells. Other shells should do this too, but let's start with fish.

### DIFF
--- a/libexec/rbenv-init
+++ b/libexec/rbenv-init
@@ -86,8 +86,10 @@ mkdir -p "${RBENV_ROOT}/"{shims,versions}
 
 case "$shell" in
 fish )
-  echo "set -gx PATH '${RBENV_ROOT}/shims' \$PATH"
-  echo "set -gx RBENV_SHELL $shell"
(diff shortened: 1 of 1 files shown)


libexec/rbenv-init
```diff
@@ -86,8 +86,10 @@ mkdir -p "${RBENV_ROOT}/"{shims,versions}
 
 case "$shell" in
 fish )
-  echo "set -gx PATH '${RBENV_ROOT}/shims' \$PATH"
-  echo "set -gx RBENV_SHELL $shell"
+  echo "if status --is-login"
```
mislav: We already do this check when giving people instructions how to set up `rbenv init` in their fish initialization file: https://github.com/rbenv/rbenv/blob/2d7cefe78279f65a6aa3c6e5db1e5daffe934061/libexec/rbenv-init#L73

Do you think that `is-interactive` check isn't enough? What does `is-login` achieve?

mislav:
Closing as wontfix

PR #1058, opened 2017-12-18
use absolute path to rbenv and check if PATH was already extended

kt97679:
I noticed that rbenv-init doesn't check if ${RBENV_ROOT}/shims is already added to the PATH, this is fixed on line 93.

In order to use rbenv I have to create symlink in $HOME/bin to ${RBENV_ROOT}/bin/rbenv. With fix on lines 148 and 150 this is not needed any more. I can have
```
eval "$(${RBENV_ROOT}/bin/rbenv init -)"
```
in my .bashrc instead of 
```
eval "$(rbenv init -)"
```
and there will be no need to create symlink.

### DIFF
--- a/libexec/rbenv-init
+++ b/libexec/rbenv-init
@@ -90,7 +90,7 @@ fish )
   echo "set -gx RBENV_SHELL $shell"
 ;;
 * )
-  echo 'export PATH="'${RBENV_ROOT}'/shims:${PATH}"'
+  echo ":${PATH}:" | grep -q ":${RBENV_ROOT}/shims:" || echo 'export PATH="'${RBENV_ROOT}'/shims:${PATH}"'
   echo "export RBENV_SHELL=$shell"
 ;;
 esac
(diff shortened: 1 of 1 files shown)


mislav:
See https://github.com/rbenv/rbenv/pull/901#issuecomment-220004280

kt97679:
I see, thanks for the clarification regarding PATH @mislav . What about 2nd change?

mislav:
> What about 2nd change?

Oh sorry, I did not understand what the 2nd change was about. I don't think that code change was necessary. In order to use rbenv you don't have to symlink `$HOME/bin` to `$RBENV_ROOT/bin/rbenv`. You simply have to configure your PATH to include `$RBENV_ROOT/bin` in it. This is explained in installation instructions.

kt97679:
I'm sorry @mislav let me give you  more details regarding the second change. Yes, you are absolutely right in your comment that I need to ensure that $RBENV_ROOT/bin/rbenv is in the PATH. To do this I create symlink in my $HOME/bin that points to the $RBENV_ROOT/bin/rbenv . I do this because I already have $HOME/bin in the PATH and I don't want to add $RBENV_ROOT/bin to the PATH for the sake of a single script. It also worth mentioning that presence of rbenv in the PATH is needed only for initialization. After that rbenv command is handled by the shell function. With the proposed change there is no need to create mentioned symlink because rbenv will work just fine being called by absolute name. Please let me know if I miss something here.

Issue #1065, opened 2018-01-26
`rbenv version-name > .ruby-version` write `system` to file instead of current version


ayqazi:
My current global Ruby version is 2.5.0
```
rbenv version
=> 2.5.0 (set by /home/xxxx/.rbenv/version)

rbenv version-name
=> 2.5.0
```

However when I try to redirect that to a file, it outputs `system`!
```
rbenv version-name > .ruby-version
cat .ruby-version
=> system
```

This seems wrong.......

I'm currently on the master branch of rbenv. Commit ID as of now is `b943955dbf10c436a8d4c7c36fd4216a34ef2b33`.

Thanks

viclim:
@ayqazi 

I created a [PR](https://github.com/rbenv/rbenv/pull/1069) to address the issue you're having.

In the meantime, use the builtin method instead.
```sh
rbenv local $(rbenv version-name)
# or simply
rbenv local 2.5.0
```

PR #1069, opened 2018-02-26
Ignore empty `.ruby-version` files

viclim:
This PR addresses https://github.com/rbenv/rbenv/issues/1065.

Command `rbenv version-name > .ruby-version` will create an empty `.ruby-version` file
before running `rbenv-version-file`. This causes `rbenv-version-file` to return an empty
string which in turn causes `rbenv-version-name` to return `system`.

Ensure file size of `.ruby-version` is non-zero as a workaround.

**This fix will cause problem for users that use empty `.ruby-version` to point to `system` ruby.**

### DIFF
--- a/libexec/rbenv-version-file
+++ b/libexec/rbenv-version-file
@@ -9,7 +9,7 @@ target_dir="$1"
 find_local_version_file() {
   local root="$1"
   while ! [[ "$root" =~ ^//[^/]*$ ]]; do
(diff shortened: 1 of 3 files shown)


mislav:
GitHub tips:
* Use a descriptive PR title instead of “Fix <URL>”. URLs in PR titles are not clickable.
* If you reference an issue by saying `fixes #1065` ("closes" would also work), then that issue will be auto-closed when this PR is merged.

PR #1075, opened 2018-04-13
Run rehash command in the background

andersonvom:
The rehash command doesn't take too long to run, but can add up when
several *env projects (rbenv, pyenv, nodenv, etc) are installed at the
same time. This makes it run in the background so that you don't have
to skip it altogether (using the `--no-rehash` flag) but also don't
have to wait a lot during shell initialization.

The `()` around the command prevent it from displaying the job number
or the "job done" message once it finishes.

### DIFF
--- a/libexec/rbenv-init
+++ b/libexec/rbenv-init
@@ -101,7 +101,7 @@ if [ -r "$completion" ]; then
 fi
 
 if [ -z "$no_rehash" ]; then
-  echo 'command rbenv rehash 2>/dev/null'
+  echo 'command rbenv rehash 2> /dev/null &'
 fi
 
 commands=(`rbenv-commands --sh`)
--- a/test/init.bats
+++ b/test/init.bats
@@ -14,7 +14,7 @@ load test_helper
 @test "auto rehash" {
   run rbenv-init -
   assert_success
-  assert_line "command rbenv rehash 2>/dev/null"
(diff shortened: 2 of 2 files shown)


andersonvom:
Bump

libexec/rbenv-init
```diff
@@ -101,7 +101,7 @@ if [ -r "$completion" ]; then
 fi
 
 if [ -z "$no_rehash" ]; then
-  echo 'command rbenv rehash 2>/dev/null'
+  echo '(command rbenv rehash 2>/dev/null &)'
```
mislav: Looks good for bash, but this is going to be eval'd on shells such as zsh, tcsh, ksh, fish, and others. Do we have any guarantees that the same syntax will work on all of these shells?

I'm reluctant to pull this because of compatibility concerns. And also, if someone wants to speed up the rehash process when starting a new shell, they should just skip it with `--no-rehash`.

Finally, you can massage `rbenv init` output yourself to add the parentheses:

```sh
eval "$(rbenv init - | sed-E 's/(.+rbenv rehash.+)/(\1 \&)/')"
```

andersonvom: You are correct. I've tested this syntax on `fish` and it does not work. However, the current syntax does not work on `tcsh`, so maybe it would be nice to have a set of supported shells.

As to using `--no-rehash`, it's a bit not ideal as you lose functionality.

PR #1076, opened 2018-04-24
Wait rehash until lock acquisition

yyuu:
rbenv and ruby-build are both implemented to call rehash inside its code on some circumstances. However, since the rehash is implemented to fail if there're another concurrent rehash process within the same `$RBENV_ROOT`, all these implementations cannot run in parallel.

In general, rehash will finish within a second for most cases. It'd mean that just adding some extra lock wait of few seconds will help improving concurrency of these rehash related commands without introducing other external lock. (AFAIK, some external tool like Jenkins' [rbenv-plugin is using some external lock](https://github.com/jenkinsci/rbenv-plugin/blob/v0.0.16/lib/rbenv/semaphore.rb#L25) as a workaround for the concurrent rehash issue)

I've implemented some retry mechanism based on current code of `noclobber` based locking. It will retry up to 60 seconds until acquisition of the _lock file_.

See also:
* https://github.com/pyenv/pyenv/pull/1145

### DIFF
--- a/libexec/rbenv-rehash
+++ b/libexec/rbenv-rehash
@@ -40,7 +40,7 @@ if [ ! -w "$SHIM_PATH" ]; then
 fi
 
 unset acquired
-for _ in $(seq "${RBENV_REHASH_TIMEOUT:-60}"); do
+for (( i=1; i<="${RBENV_REHASH_TIMEOUT:-60}"; i++ )); do
   if acquire_lock 2>/dev/null; then
     acquired=1
     break
--- a/libexec/rbenv-rehash
+++ b/libexec/rbenv-rehash
@@ -10,29 +10,50 @@ PROTOTYPE_SHIM_PATH="${SHIM_PATH}/.rbenv-shim"
 # Create the shims directory if it doesn't already exist.
 mkdir -p "$SHIM_PATH"
 
-# Ensure only one instance of rbenv-rehash is running at a time by
-# setting the shell's `noclobber` option and attempting to write to
-# the prototype shim file. If the file already exists, print a warning
-# to stderr and exit with a non-zero status.
-set -o noclobber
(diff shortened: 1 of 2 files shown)


mislav:
Hi, thanks for the contribution. In which cases would this help? Also, why would we ever allow `rbenv rehash` to run up to 60s?

yyuu:
There are some internal calls of `rbenv-rehash` within rbenv, and they are expecting `rbenv-rehash` will _less likely fail_ (at least they won't retry after any failure).

https://github.com/rbenv/rbenv/blob/v1.1.1/libexec/rbenv-init#L104
https://github.com/rbenv/rbenv/blob/v1.1.1/rbenv.d/exec/gem-rehash/rubygems_plugin.rb#L30

Due to the missing appropriate failure handling within these code, concurrent executions of some processes (e.g. start up multiple shell with rbenv at a time, run multiple `gem install` at a time, etc.) will cause some obscure failures. This might not  be typical for single user use case, but would not be different for something like CI build environment (and I know rbenv is useful for these environments).

> Also, why would we ever allow `rbenv rehash` to run up to 60s?

I guessed that the most typical use case of this feature is some automated build environments, and I also guessed that 1 minute delay must not be significant for such environments. Although it's kinda overkill and using some more shorter value (e.g. `5s`, `10s`) as the default would work well for most typical use cases. I can tweak the default value if you prefer.

mislav:
Thank you so much for developing this solution. The code looks great, but I'm afraid I won't pull this into rbenv because I'm trying to reduce the amount of functional changes I make to rbenv these days, and I just don't have the evidence that concurrent rehash is a huge problem with users of rbenv. For those mechanisms such as Jenkins, I'm fine with a lock being acquired externally from rbenv.

Issue #865, opened 2016-02-02
rbenv-exec only allows executables found by rbenv-which


jasonkarns:
I don't think this is really a bug but more of a feature request. Presently, `rbenv exec` doesn't do what is advertised on the cover. Per the help:

```
Runs an executable by first preparing PATH so that the selected Ruby
version's `bin' directory is at the front.

For example, if the currently selected Ruby version is 1.9.3-p327:
  rbenv exec bundle install

is equivalent to:
  PATH="$RBENV_ROOT/versions/1.9.3-p327/bin:$PATH" bundle install
```

To me, this means that `rbenv exec foo` should be the same as `PATH="$RBENV_ROOT/versions/1.9.3-p327/bin:$PATH" foo`. But instead, rbenv-exec checks rbenv-which first. So if the provided command doesn't exist under the currently selected ruby, then we get an error rather than simply invoking the command.

I'm not sure if this behavior belongs in rbenv-exec (though that's my inclination). But it should be possible to simply execute a command while letting rbenv set up the correct `PATH` for me first.

mislav:
`rbenv-exec` was solely designed for powering the shims mechanism, and for that purpose it makes sense to focus only on executables that are in ruby bin directory. However I see your point re: manual usage. Maybe `rbenv-exec` should have 2 operation modes?

jasonkarns:
I just ran into this issue again. the rbenv-env plugin is intended to print out useful env vars relevant to rbenv (that is, env vars such as `GEM*`, `RUBY*`, `RAILS*`, `RBENV_*`, etc). However, in order to print the `PATH` as it exists in the context of rbenv; it has to manually do all of the `PATH` setup that `rbenv-exec` does. It essentially re-implements `rbenv-exec`, including running any `exec` hooks such that any environment variable modifications done by `exec` hooks are respected. (In addition to manually modifying `PATH` such that `PATH` is printed as it would exist under rbenv.)

This is clearly suboptimal. In reality, the ideal implementation of rbenv-env would simply be `rbenv exec env` (Run `env` in the context of rbenv) but of course that fails because `env` is not a ruby command.

Do you have any thoughts on how this should operate?

jasonkarns:
Related: #187

hurricup:
Any updates on this? Would be really nice to attempt to execute as is if no shim exists.
Or, even make some thing like `rbenv 2.5.1 run arbitrary_command`

PR #1082, opened 2018-05-26
exec --no-resolve parameter introduced

hurricup:
Allows to execute arbitrary command in the context of selected ruby.
#865


hurricup:
Another way is to adjust `rbenv-which` behavior, so it could resolve absolute paths as fallback. But this looks more explicit to me.

Qwert9090aa (approved):
libexec/rbenv-exec

Issue #459, opened 2013-10-05
rbenv shell command not found


mdesantis:
Environment:

```
$ uname -srvm
Linux 3.11.0-11-generic #17-Ubuntu SMP Tue Oct 1 19:42:04 UTC 2013 x86_64
$ echo $SHELL
/usr/bin/fish
$ rbenv --version
rbenv 0.4.0-65-gff23666
```

When I execute `rbenv shell jruby-1.7.4` rbenv outputs <code>rbenv: no such command `shell'</code> and the ruby environment doesn't change.

In order to mitigate this issue I'm using this:

```
function rbenv-shell
  set --local RUBY_VERSION $argv[1]

  if test "$RUBY_VERSION" = "--unset"
    set --global --export --erase RBENV_VERSION
  else if test -n "$RUBY_VERSION"
    set --global --export RBENV_VERSION $RUBY_VERSION
  end

  rbenv version
end
complete --command rbenv-shell --exclusive --arguments "--unset (rbenv versions --bare)"
```

chiqui3d:
I solved it in ubuntu by installing coreutils again

Issue #195, opened 2012-02-16
Shims for Fish shell


gf3:
It would be very helpful is rbenv provided shims that were also fish shell compatible. It wouldn't be a huge change, but it'd mean that rbenv could be used in a non-bash/zsh shell. The syntax for fish is actually much more sane than bash.

muloka:
Thanks @emcmanus and @sstephenson.

ChFlick:
Apparently the ^ syntax will be deprecated in the future and is broken in some versions ([see this issue here](https://github.com/fish-shell/fish-shell/issues/6591#issuecomment-585775178)).

Therefore it may be better to use the following syntax instead: `rbenv rehash >/dev/null >2&1`.

JelteF:
> Therefore it may be better to use the following syntax instead: rbenv rehash >/dev/null >2&1.

There's a typo in there `2` and `>` should be swapped:
```bash
rbenv rehash >/dev/null 2>&1
```

PR #1083, opened 2018-05-30
Update instructions for $PATH change in README

AlexWayfer:
Describe for Fish shell in details, prevent issues like #195 and #459.

Reference: https://fishshell.com/docs/2.7/tutorial.html#tut_path


mislav:
Thank you!

PR #1085, opened 2018-05-31
libexec/rbenv: skip `cd`s

blueyed:
This appears to be done to get the resolved name for `$RBENV_DIR`, but
that is not really necessary.

This adds an explicit existence check instead, which also gets done for
`$PWD`, because it might not exist after all, too.

### DIFF
--- a/libexec/rbenv
+++ b/libexec/rbenv
@@ -61,9 +61,9 @@ if [ -z "${RBENV_DIR}" ]; then
   RBENV_DIR="$PWD"
 else
   [[ $RBENV_DIR == /* ]] || RBENV_DIR="$PWD/$RBENV_DIR"
-  cd "$RBENV_DIR" 2>/dev/null || abort "cannot change working directory to \`$RBENV_DIR'"
-  RBENV_DIR="$PWD"
-  cd "$OLDPWD"
+fi
+if [ ! -d "$RBENV_DIR" ] || [ ! -e "$RBENV_DIR" ]; then
+  abort "cannot change working directory to \`$RBENV_DIR'"
 fi
 export RBENV_DIR
 


mislav:
The manual expansion of RBENV_DIR was added intentionally 7fbcde8af0e7ac0afdeba21b5f7a5d37fc3faa93 and it only kicks in if RBENV_DIR was provided explicitly via environment variable. This behavior was added to accommodate the possibly relative RBENV_DIR value provided by shims: https://github.com/rbenv/rbenv/blob/bc1b66250f564e25409e5300391b4c5612db0d57/libexec/rbenv-rehash#L55

We haven't had a problem with it, so I don't see why it should change.

blueyed:
It was mainly meant to avoid some overhead.
As for relative paths they are made absolute to $PWD already.
So the only benefit of the current code is to resolve symlinks, but that is not really necessary, is it?

blueyed:
btw: pyenv even has this additionally: https://github.com/pyenv/pyenv/pull/1164/files#diff-414cbe2ca9ee5fbfdc48e82392f82db7R84.

mislav:
It's not about expanding symlinks in this case, it's simply about absolutizing the path because in the subsequent rbenv logic, RBENV_DIR is treated as absolute: https://github.com/rbenv/rbenv/blob/bc1b66250f564e25409e5300391b4c5612db0d57/libexec/rbenv-version-file#L25-L26

blueyed:
That is covered by https://github.com/rbenv/rbenv/pull/1085/files#diff-96f9c843ac1f142eca3a25b2e3db090fR63 then, isn't it?  (although it only handles POSIX paths (i.e. looks if it starts with a slash, which might be an issue on Windows?!))

mislav:
It's not exactly covered by that, since RBENV_DIR might have `.` and `..` in it, and we want to expand those into an absolute path. The easiest way to do that is with a `cd .. $PWD` combo.

PR #1087, opened 2018-06-04
libexec/rbenv-exec: simplify/fix sourcing exec hooks

blueyed:
Fixes two issues reported by shellcheck:

    In libexec/rbenv-exec line 37:
    IFS=$'\n' scripts=(`rbenv-hooks exec`)
                       ^-- SC2207: Prefer mapfile or read -a to split command output (or quote to avoid splitting).
                       ^-- SC2006: Use $(..) instead of legacy `..`.

### DIFF
--- a/libexec/rbenv-exec
+++ b/libexec/rbenv-exec
@@ -33,10 +33,7 @@ export RBENV_VERSION
 RBENV_COMMAND_PATH="$(rbenv-which "$RBENV_COMMAND")"
 RBENV_BIN_PATH="${RBENV_COMMAND_PATH%/*}"
 
-OLDIFS="$IFS"
-IFS=$'\n' scripts=(`rbenv-hooks exec`)
-IFS="$OLDIFS"
-for script in "${scripts[@]}"; do
+rbenv-hooks exec | while read -ra script; do
   source "$script"
 done
 
--- a/libexec/rbenv-init
+++ b/libexec/rbenv-init
@@ -1,6 +1,7 @@
 #!/usr/bin/env bash
 # Summary: Configure the shell environment for rbenv
 # Usage: eval "$(rbenv init - [--no-rehash] [<shell>])"
+# shellcheck disable=SC2016,SC2086,SC2088
 
 set -e
 [ -n "$RBENV_DEBUG" ] && set -x
@@ -104,7 +105,10 @@ if [ -z "$no_rehash" ]; then
   echo 'command rbenv rehash 2>/dev/null'
 fi
 
-commands=(`rbenv-commands --sh`)
+shell_commands=()
+while read -ra shell_command; do
+  shell_commands+=("$shell_command")
+done < <(rbenv-commands --sh)
 case "$shell" in
 fish )
   cat <<EOS
@@ -113,7 +117,7 @@ function rbenv
   set -e argv[1]
 
   switch "\$command"
-  case ${commands[*]}
+  case ${shell_commands[*]}
     source (rbenv "sh-\$command" \$argv|psub)
   case '*'
     command rbenv "\$command" \$argv
@@ -144,7 +148,7 @@ cat <<EOS
   fi
 
   case "\$command" in
-  ${commands[*]})
+  ${shell_commands[*]})
     eval "\$(rbenv "sh-\$command" "\$@")";;
   *)
     command rbenv "\$command" "\$@";;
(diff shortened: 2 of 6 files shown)


libexec/rbenv-exec
```diff
@@ -33,10 +33,7 @@ export RBENV_VERSION
 RBENV_COMMAND_PATH="$(rbenv-which "$RBENV_COMMAND")"
 RBENV_BIN_PATH="${RBENV_COMMAND_PATH%/*}"
 
-OLDIFS="$IFS"
-IFS=$'\n' scripts=(`rbenv-hooks exec`)
-IFS="$OLDIFS"
```
mislav: Why no restoring original IFS value?

blueyed: It gets used only for the current command, but that applies to `rbenv-hooks`.
But it's also not necessary, since `read` splits by newlines already.

blueyed:
Amended.

blueyed:
btw: (why) is Travis not running tests?

libexec/rbenv-exec
mislav: The value of IFS affects hook scripts that get eval'd in the current scope. It would be good for IFS to be at the default value, otherwise some scripts that depend on word splitting might break.

libexec/rbenv-exec
```diff
@@ -33,10 +33,7 @@ export RBENV_VERSION
...
-for script in "${scripts[@]}"; do
+rbenv-hooks exec | while read -ra script; do
```
mislav: I was trying to remember why I didn't go the pipe-to-while loop approach, since this looks much cleaner. Now I remember. It has to do with variable scopes:

```sh
RBENV_FOO=init
printf 'foo\nbar\n' | while read -ra line; do
  echo "read: $line"
  RBENV_FOO="$line"
done

echo $RBENV_FOO #=> "init"
```

We expected RBENV_FOO to change, but it didn't. It's imperative that hook scripts have the ability to add and change environment variables. To allow this, I think we can't afford to pipe to a loop.

mislav: Also (this is minor): if the input piped to `while read` doesn't have the terminating newline character, the last line won't be read. This might not affect the `rbenv-hooks COMMAND | while read` combo specifically, but is another thing to be aware of and why splitting with IFS might sometimes be more effective.

Also, why would you `read -a`? It doesn't seem like `-a` is desired here

blueyed: Ah, then we can pipe into the loop:

```sh
RBENV_FOO=init
while read -ra line; do
  echo "read: $line"
  RBENV_FOO="$line"
done < <(printf 'foo\nbar\n')

echo $RBENV_FOO #=> "bar"
```

libexec/rbenv-exec
blueyed: What I mean is that `IFS=:` only applies to `foo` in `IFS=: foo; bar`, and not to `bar` (AFAIK).
It is different from `IFS=:; foo; bar` (3 commands), vs 2 commands.

libexec/rbenv-exec
mislav: We avoid process substitution syntax `<(...)` because it caused problems for some people: 681435e0b4a0d3f9609a11af734902312eabaeaf

libexec/rbenv-exec
mislav: That's usually true, but not in case of assignment.

```sh
OLDIFS="$IFS"
IFS=$'\n' words=( $(printf 'foo bar\nbaz\n') )
echo "${#words[@]}" #=> 2

words=( $(printf 'foo bar\nbaz\n') )
echo "${#words[@]}" #=> 2; expected 3

IFS="$OLDIFS"
words=( $(printf 'foo bar\nbaz\n') )
echo "${#words[@]}" #=> 3
```

It would be good if you tested your claims before trying to convince me of them 😄Believe me, I tried many of these approaches during many sleepless nights.

I'm open to accepting a PR that fixes this but *doesn't* break compatibility in the process.

libexec/rbenv-exec
blueyed: Oh snap.. :D

What is the state of https://github.com/mislav/anyenv ? :)

mislav: I haven't worked on it since the last commit activity that you can see (a long time ago!) but I would be interested in wrapping it up and making it shippable!

libexec/rbenv-exec
blueyed: Yeah, but I've never used it in this PR before an assignment, have I? :)

blueyed:
Closing.

PR #1088, opened 2018-06-04
rbenv-exec: do not add RBENV_BIN_PATH to PATH

blueyed:
This appears to have been added for "ruby -S" support (which looks up
the script name in $PATH).
But it looks like this should just get delegated to "ruby" itself.

Initially I've looked into looking for the file in the shim (when "-S"
is used), but it does not seem to be necessary after all.

I've looked into this when trying to fix the issues in pyenv that arise
from mangling the $PATH here.

### DIFF
--- a/libexec/rbenv-exec
+++ b/libexec/rbenv-exec
@@ -41,7 +41,4 @@ for script in "${scripts[@]}"; do
 done
 
 shift 1
-if [ "$RBENV_VERSION" != "system" ]; then
-  export PATH="${RBENV_BIN_PATH}:${PATH}"
-fi
 exec -a "$RBENV_COMMAND" "$RBENV_COMMAND_PATH" "$@"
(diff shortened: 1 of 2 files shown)


blueyed:
I've did some manual testing using `PATH=$PWD/t:$PATH ruby -S t.rb` with a file `t/t.rb` and it worked as expected.

blueyed:
pyenv issue: https://github.com/pyenv/pyenv/pull/1169

This also uses the full path with `exec -a`, which appears to be required for `python` - not sure about `ruby`.
Is there an equivalent of `sys.executable` that gets derived from `argv[0]`, similar to how Python does it?  (it looks up argv0 in the $PATH if it is not absolute)

blueyed:
Included in https://github.com/rbenv/rbenv/pull/1089.

PR #1089, opened 2018-06-04
Fix PATH mangling and do not use -a with exec

blueyed:
This stores the original $PATH in RBENV_ORIG_PATH and uses that when
`exec`ing the actual command.

Includes #1088 and https://github.com/rbenv/rbenv/pull/1090.

### DIFF
--- a/libexec/rbenv
+++ b/libexec/rbenv
@@ -71,10 +71,11 @@ export RBENV_DIR
 shopt -s nullglob
 
 bin_path="$(abs_dirname "$0")"
+export _RBENV_ORIG_PATH="$PATH"
 for plugin_bin in "${RBENV_ROOT}/plugins/"*/bin; do
   PATH="${plugin_bin}:${PATH}"
 done
-export PATH="${bin_path}:${PATH}"
+PATH="${bin_path}:${PATH}"
 
 RBENV_HOOK_PATH="${RBENV_HOOK_PATH}:${RBENV_ROOT}/rbenv.d"
 if [ "${bin_path%/*}" != "$RBENV_ROOT" ]; then
--- a/libexec/rbenv-exec
+++ b/libexec/rbenv-exec
@@ -31,7 +31,6 @@ fi
 
 export RBENV_VERSION
 RBENV_COMMAND_PATH="$(rbenv-which "$RBENV_COMMAND")"
-RBENV_BIN_PATH="${RBENV_COMMAND_PATH%/*}"
 
 OLDIFS="$IFS"
 IFS=$'\n' scripts=(`rbenv-hooks exec`)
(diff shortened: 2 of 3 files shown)


libexec/rbenv
```diff
@@ -71,10 +71,11 @@ export RBENV_DIR
 shopt -s nullglob
 
 bin_path="$(abs_dirname "$0")"
+export RBENV_ORIG_PATH="$PATH"
```
blueyed: Prefix it with `_` maybe?

blueyed: exporting it also is quite bad (should be unset probably at least), but necessary due to rbenv using subshells for its own programs.

blueyed: (see https://github.com/pyenv/pyenv/pull/357 and https://github.com/pyenv/pyenv/pull/1153 for not having to `export` this, but it shouldn't hold this back)

mislav:
Sorry, but this fundamentally changes how rbenv works and I think the change would be too disruptive.

The whole design system of rbenv is that it pushes the current Ruby version to PATH so that all the child processes hit the executables of the currently selected Ruby immediately. Changing that would cause all sorts of backwards-incompatiblity issues.

Also, `exec -a` didn't prove to be harmful for our purposes so far, so I'm keeping it. It keeps the value of `$0` clean.

blueyed:
Oh, too bad.

What about https://github.com/rbenv/rbenv/pull/1089/commits/4fe1075e7d4d96f3bfee88378134cb5567784d93? (not adding libexec and plugin bins to $PATH)

I've looked into using `RBENV_PATH` instead (which would also include $PATH), and not change $PATH itself.

> The whole design system of rbenv is that it pushes the current Ruby version to PATH so that all the child processes hit the executables of the currently selected Ruby immediately.

But wouldn't they go through the shims then also otherwise?
So it would be only for performance then?
(I have some `pyenv-prefix-to-path` (adding `pyenv prefix` to the beginning of $PATH to skip shims), which seems to be similar)

btw: there is also some confusion created by the fact that a provided script triggers the Python/Ruby version being used, but that is something different, right?  https://github.com/pyenv/pyenv/issues/1163#issuecomment-393525105

As for the `exec -a` change, I will push this for pyenv then already.

blueyed:
> Rbenv wants to initially execute as quickly as possible and then move itself out of the way.

Makes sense.

> Why are you so insterested in keeping PATH intact, if I may ask?

I think the main issue with `pyenv` is that it also changes $PATH for the system version, which was added due to the `exec -a` to fix https://github.com/pyenv/pyenv/issues/98.
This is not necessary anymore when dropping that, and not changing $PATH for the system version.

But in general I think it is good to not mangle the $PATH (which also adds additional overhead for subprocesses).

> If you wanted to open an exploratory PR how this could be achieved in rbenv (simple solutions are preferred) I would be interested in reviewing it!

Great.

PR #1090, opened 2018-06-04
rbenv-exec: do not use `-a` with `exec`

blueyed:
Using the command name only causes an issue in pyenv, where Python then
uses the PATH to look up the full path for `sys.executable`.

Not using `-a` appears to result in the same behavior as using the full
command path (`$RBENV_COMMAND_PATH`), so just do not use it.
It has been there since the initial commit (99035a4).


blueyed:
Included in https://github.com/rbenv/rbenv/pull/1089 also, with conflicts resolved already.

PR #1091, opened 2018-06-04
rehash: shims: source rbenv-exec

blueyed:
This avoids an unnecessary subshell for `command -v`, and sources
libexec/rbenv-exec instead of calling it through a new bash subprocess.


mislav:
Sorry, but I don't see the value in this change. A simple subshell to `command -v` won't significantly impact `rbenv rehash` performance.

PR #1092, opened 2018-06-07
make rbenv init more visible in the readme

jacobherrington:
In these two places where the reader is instructed to run `rbenv init` the command is not broken out from the text like nearly all of the other shell commands. This makes it easy for a reader to miss or misunderstand that `rbenv init` should be run in shell.

I moved them into a position such that they are now harder to miss and consistent with the rest of the document. Let me know if there are other changes I should make.

Thanks.

### DIFF
--- a/README.md
+++ b/README.md
@@ -169,9 +169,7 @@ If you're on macOS, we recommend installing rbenv with
     $ rbenv init
     ~~~
 
-   Follow the printed instructions to [set up rbenv shell integration](#how-rbenv-hooks-into-your-shell). 
-   This is the step that will make running `ruby` "see" the Ruby version that 
-   you choose with rbenv.
+   Follow the printed instructions to [set up rbenv shell integration](#how-rbenv-hooks-into-your-shell).
 
 3. Close your Terminal window and open a new one so your changes take
    effect.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -163,9 +163,15 @@ If you're on macOS, we recommend installing rbenv with
    Note that this also installs `ruby-build`, so you'll be ready to
    install other Ruby versions out of the box.
 
-2. Run `rbenv init` and follow the instructions to set up
-   rbenv integration with your shell. This is the step that will make
-   running `ruby` "see" the Ruby version that you choose with rbenv.
+2. Initialize rbenv.
```
mislav: “Initialize rbenv” says little on its own. How about “Set up rbenv in your shell”?

README.md
```diff
@@ -163,9 +163,15 @@ If you're on macOS, we recommend installing rbenv with
...
+
+    ~~~ sh
+    $ rbenv init
+    ~~~
+
+   Follow the instructions to set up rbenv integration with your shell. This
```
mislav: I know that you mostly preserved old text, but let's take this opportunity to improve it. How about:

```
Follow the printed instructions to [set up rbenv shell integration](#how-rbenv-hooks-into-your-shell).
```

jacobherrington:
@mislav made those changes, thanks for the suggestions!

PR #1093, opened 2018-06-07
rbenv-which: change PATH only for the "command -v" lookup

blueyed:
This is not really necessary, because rbenv-which is used in a subshell
currently, but makes a difference if rbenv-which would be sourced.


mislav:
Good catch; thanks!

PR #1098, opened 2018-07-13
rbenv-prefix: do not silence rbenv-which for system version

blueyed:
This suppressed any output when using RBENV_DEBUG=1 and does not really
hurt to have in the unlikely case that it should fail; you would get
two error messages now:

rbenv: ruby: command not found
rbenv: system version not found in PATH

### DIFF
--- a/libexec/rbenv-prefix
+++ b/libexec/rbenv-prefix
@@ -22,7 +22,7 @@ elif [ -z "$RBENV_VERSION" ]; then
 fi
 
 if [ "$RBENV_VERSION" = "system" ]; then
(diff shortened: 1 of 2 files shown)


mislav:
Thanks, but I think one error message is enough! This is a rare case anyway.

blueyed:
> Thanks, but I think one error message is enough!

I think the main issue here is that it also silences all of `RBENV_DEBUG`.

mislav:
> I think the main issue here is that it also silences all of `RBENV_DEBUG` (for the subprocess).

Ah I see. That's a fair point. I'm willing to accept a PR that doesn't break the test suite.

blueyed:
Thanks, tests should be passing now.

PR #1099, opened 2018-07-13
Improve init: warn about missing shell and use it in the template

blueyed:
Providing the shell makes rbenv-init "only" take 50ms, instead of 90ms.

The slow part is coming from the `ps` that runs in a subshell when the
shell is not provided.

Therefore this will print a warning, and updates the template that
`rbenv init` itself generates.

I think in the long run `shell` should become a required argument, but
this is a good first step already.

### DIFF
--- a/libexec/rbenv-init
+++ b/libexec/rbenv-init
@@ -71,10 +71,10 @@ if [ -z "$print" ]; then
     echo
     case "$shell" in
     fish )
-      echo 'status --is-interactive; and rbenv init - | source'
+      echo 'status --is-interactive; and rbenv init - fish | source'
       ;;
     * )
-      echo 'eval "$(rbenv init -)"'
+      printf 'eval "$(rbenv init - %s)"\n' "$shell"
       ;;
     esac
     echo
--- a/test/init.bats
+++ b/test/init.bats
@@ -51,10 +51,16 @@ OUT
   [ -z "$line" ] || flunk "did not expect line: $line"
 }
 
+@test "posix shell instructions" {
+  run rbenv-init bash
+  assert [ "$status" -eq 1 ]
+  assert_line 'eval "$(rbenv init - bash)"'
+}
+
 @test "fish instructions" {
   run rbenv-init fish
   assert [ "$status" -eq 1 ]
-  assert_line 'status --is-interactive; and rbenv init - | source'
+  assert_line 'status --is-interactive; and rbenv init - fish | source'
 }
 
 @test "option to skip rehash" {


libexec/rbenv-init
```diff
@@ -70,10 +73,10 @@ if [ -z "$print" ]; then
     echo
     case "$shell" in
     fish )
-      echo 'status --is-interactive; and source (rbenv init -|psub)'
+      echo 'status --is-interactive; and source (rbenv init - fish|psub)'
       ;;
     * )
-      echo 'eval "$(rbenv init -)"'
+      printf 'eval "$(rbenv init - %s)"' "$shell"
```
mislav: Interesting idea, but I only see an improvement of 4ms at most, which I don't think is worth it. Is there something about your system that makes `ps -p $PPID` take 40ms?

blueyed: Measuring it directly is a bit less, but still quite a bit for something unnecessary:

```
% python -m timeit -s 'import subprocess' -- 'subprocess.Popen(["ps", "-p", "15368"], stdout=subprocess.DEVNULL).communicate()'
20 loops, best of 5: 13.3 msec per loop

% python -m timeit -s 'import subprocess' -- 'subprocess.Popen(["bash", "-c", "ps -p 15368"], stdout=subprocess.DEVNULL).communicate()'
20 loops, best of 5: 15.9 msec per loop
```

blueyed: A bit more even when using the actual args:
```
% python -m timeit -s 'import subprocess' -- 'subprocess.Popen(["bash", "-c", "ps -p 15386 -o args="], stdout=subprocess.PIPE).communicate()'                               
10 loops, best of 5: 19.6 msec per loop
```

mislav: You're including the startup time of bash in your measurement. For me, this consistently takes 0.004s:

```sh
TIMEFORMAT='%3R'
time ps -p $$ -o args=
```

blueyed: Well, shelling out is being done there after all.

What I've used inititally for timing was `RBENV_ROOT=$PWD timeit-shell bin/rbenv init -` vs `RBENV_ROOT=$PWD timeit-shell bin/rbenv init - zsh`.

Where timeit-shell is:
```
#!/usr/bin/python

import sys
import timeit

timeit.main(['--setup', 'import subprocess', '--',
             'subprocess.Popen(%r, stdout=subprocess.PIPE).communicate()' % (sys.argv[1:],)])
```

You can use this on master, since it only shows the difference of using the shell or calling `ps`.

For me it is:
```
% RBENV_ROOT=$PWD timeit-shell bin/rbenv init - zsh
10 loops, best of 5: 21.4 msec per loop

% RBENV_ROOT=$PWD timeit-shell bin/rbenv init
10 loops, best of 5: 40.8 msec per loop
```

So it is ~twice as fast.
And the overhead (caused by the `ps`) appears to be ~20ms.
And this pretty much matches the initial python-only measuring!

From your example I am getting ~0.032-0.037 btw - and I should have a very stripped down bash config, since I am using Zsh normally.  This is a Lenovo x250 with an SSD.

blueyed:
I have no idea btw to adopt this for fish.. :/

libexec/rbenv-init
blueyed: btw: please do not wear me out on this: I am not using rbenv (but only pyenv occasionally, but then also not pyenv-init by itself), but only try to improve things upstream (which includes pyenv already for me, since I am not using its `init`).

mislav: > btw: please do not wear me out on this: I am not using rbenv (but only pyenv occasionally, but then also not pyenv-init by itself), but only try to improve things upstream

And I thank you for your contributions! But I also want each update to rbenv to have a meaningful impact, so I want to be sure that we're improving things for everyone and not just on a certain system. That's why I'm trying to understand why `ps` is so fast for me and so slow for you. Are you using `/bin/ps`?

I tried your `timeit-shell` script and, on my macOS, your optimization reduces the time from ~41.8ms to ~38.6ms. Not a huge improvement.

I'd be open to potentitally pulling this if there is evidence it has a greater impact on other people's systems (you've provided evidence from your own end; thank you!), but I would like it to pass CI checks please.

G-Rath:
Just to chuck some two cents in:

I was looking into why the sourcing of my `.bashrc` was taking ~1s on WSLv1 (Ubuntu 18.04).

Relatively speaking, one of the two "slowest" things I found was `$(rbenv init -)` & `$(nodenv init -)`, which would both take between ~0.19s & ~0.21s.

I can confirm that just by adding '%s' to the end of `rbenv init -` takes it to ~0.18s

It's obviously not a huge boost, but compared to pretty everything else taking ~0.01s, it's nice to have a bit more speed :)

PR #1101, opened 2018-07-15
Speed up shims

marcthe12:
Solves #1100

PR #1109, opened 2018-08-09
Move the $RBENV-DIR functionality to an rbenv exec

marcthe12:
Replaces #1101. This patch tries to move shim functionaly from shims to rbenv exec since this should be universal, ie no diffence between rbenv exec $COMMAND and $COMMAND

### DIFF
--- a/libexec/rbenv-exec
+++ b/libexec/rbenv-exec
@@ -29,6 +29,20 @@ if [ -z "$RBENV_COMMAND" ]; then
   exit 1
 fi
 
+if [ "\$program" = "ruby" ]; then
+  for arg; do
+    case "\$arg" in
+    -e* | -- ) break ;;
+    */* )
+      if [ -f "\$arg" ]; then
+        export RBENV_DIR="\${arg%/*}"
+        break
+      fi
+      ;;
+    esac
+  done
+fi
+
 export RBENV_VERSION
 RBENV_COMMAND_PATH="$(rbenv-which "$RBENV_COMMAND")"
 RBENV_BIN_PATH="${RBENV_COMMAND_PATH%/*}"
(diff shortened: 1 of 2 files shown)


mislav:
Hi thank you for your contributions. What does this gain us? I'm not entirely clear on what are these series of PRs fixing.

marcthe12:
Well there should be no difference between the binstub and rbenv exec. For example, ruby /path/to/script.rb and rbenv exec ruby /path/to/script.rb invoke different version of ruby if /path/to/.ruby-version is different to local version. After all shims should be just an rbenv exec $COMMAND nothing more.

mislav:
> Well there should be no difference between the binstub and rbenv exec.

That might be your opinion, but is not a defined contract within this project. Tell me more about why are you using `rbenv exec` manually? Typical uses of rbenv should just invoke binstubs (i.e. `ruby ...`) instead of interfacing with `rbenv exec ruby ...`. `rbenv-*` commands should mostly only ever be used from plugins & other tools that for some reason need to work explicitly with rbenv.

marcthe12:
> Tell me more about why are you using rbenv exec manually?

Well a few time I want use an version of an interactive interpreter (well never in ruby per say but in python/node) to check something but not active using it. So in these cases I may not even want to rbenv init or have the shim in path. I just use rbenv exec to get that thing done.

> That might be your opinion, but is not a defined contract within this project

It is my opinion. But wont that be benefical to development as the shims in a way a frontend to rbenv in general

mislav:
Thank you for explaining! After a long deliberation, I have decided that this change isn't compelling for me to make, so I'm going to close this. Recently, I'm trying to reduce the number of functional changes made to rbenv unless absolutely necessary.

My advice would be for the caller of `rbenv exec`: if you call it manually, then you should also set RBENV_DIR manually as needed.

PR #1111, opened 2018-08-13
Sort versions semantically in rbenv versions

### LINKED ISSUE #1086: Output of `rbenv versions` not semantically sorted
At the moment, the output of `rbenv versions` is nondeterministically _(but most likely lexically)_ sorted, but it would make more sense to have it semantically sorted using the precedence rules defined by [the Semantic Versioning 2.0.0 spec](https://semver.org/).

e.g. in the following example `2.1.10` is listed before `2.1.9` even though semantically it is a later version;  the same for `2.2.10` which is listed before all other `2.2.x` versions I have installed even though it is a newer RMI version.

```shell
$ rbenv versions
  system
  2.0.0-p451
  2.1.10
  2.1.9
  2.2.10
  2.2.3
  2.2.8
  2.2.9
  2.3.3
  2.3.6
  2.3.7
  2.4.0
  2.4.2
  2.4.3
  2.4.4
* 2.5.0 (set by /usr/local/var/rbenv/version)
  2.5.1
$ _
```

A before-and-after diff would be:

```diff
--- rbenv.versions.before	2018-06-04 09:44:10.000000000 +0100
+++ rbenv.versions.after	2018-06-04 09:59:07.000000000 +0100
@@ -1,19 +1,19 @@
 $ rbenv versions
   system
   2.0.0-p451
-  2.1.10
   2.1.9
-  2.2.10
+  2.1.10
   2.2.3
   2.2.8
   2.2.9
+  2.2.10
   2.3.3
   2.3.6
   2.3.7
   2.4.0
   2.4.2
   2.4.3
   2.4.4
 * 2.5.0 (set by /usr/local/var/rbenv/version)
   2.5.1
 $ _
```

prrrnd:
A proposal to fix #1086

**Before**
```bash
$ rbenv versions
  system
  2.2.10
  2.2.5
  2.2.9
* 2.5.1 (set by /Users/prrrnd/.rbenv/version)
```

**After**
```bash
$ rbenv versions
  system
  2.2.5
  2.2.9
  2.2.10
* 2.5.1 (set by /Users/prrrnd/.rbenv/version)
```

### DIFF
--- a/libexec/rbenv-versions
+++ b/libexec/rbenv-versions
@@ -86,21 +86,36 @@ print_version() {
   num_versions=$((num_versions + 1))
 }
 
+sort_versions() {
+  sed 'h; s/[+-]/./g; s/.p\([[:digit:]]\)/.z.\1/; s/$/.z/; G; s/\n/ /' | \
+    LC_ALL=C sort -t. -k 1,1 -k 2,2n -k 3,3n -k 4,4n -k 5,5n | awk '{print $2}'
+}
+
 # Include "system" in the non-bare output, if it exists
 if [ -n "$include_system" ] && RBENV_VERSION=system rbenv-which ruby >/dev/null 2>&1; then
   print_version system
 fi
 
 shopt -s nullglob
+
+versions=($(
 for path in "$versions_dir"/*; do
   if [ -d "$path" ]; then
     if [ -n "$skip_aliases" ] && [ -L "$path" ]; then
       target="$(realpath "$path")"
       [ "${target%/*}" != "$versions_dir" ] || continue
     fi
-    print_version "${path##*/}"
+    echo "${path##*/}"
   fi
 done
+))
+
+sorted_versions=($(printf "%s\n" ${versions[@]} | sort_versions))
+
+for version in ${sorted_versions[@]}; do
+  print_version $version
+done
+
 shopt -u nullglob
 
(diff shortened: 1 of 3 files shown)


libexec/rbenv-versions
```diff
@@ -92,15 +92,29 @@ if [ -n "$include_system" ] && RBENV_VERSION=system rbenv-which ruby >/dev/null
...
     fi
-    print_version "${path##*/}"
+    echo "${path##*/}"
   fi
 done
+))
+
+previous_ifs=$IFS
+IFS=$'\n'
+versions=($(printf "%s\n" ${versions[@]} | sort -V ))
```
mislav: `sort -V` isn't available on all platforms. As far as I'm aware, it's only a feature of BSD `sort` and works on macOS and not linux.

mislav: See how we sort versions in ruby-build: https://github.com/rbenv/ruby-build/blob/00490d341bd2ee3684606397980fe3d82324a0e3/bin/ruby-build#L1182-L1185

prrrnd:
@mislav Thanks for the quick review! I'll have a look at Travis when I get some time. Any idea on how to avoid the duplication of `sort_version` from ruby-build?

mislav:
Duplication is fine.

Note that I will be away the rest of this week. Final review next week. Thank you for your contribution!

mislav (approved):

PR #1119, opened 2018-09-19
test/run: handle optional paths/args for bats

blueyed:
This allows to run a single test file more easily, although that could be
done by calling bats directly.


hsbt:
@blueyed Sorry, too late response. It's useful for me too.

PR #1140, opened 2019-01-10
Update README.md - small correction

28mm:
instructions to build the bash extension missing "./" prefix for the configure step.


mislav:
Does the command not work without the `./` prefix?

28mm:
It does. Closing.

PR #1150, opened 2019-03-05
buffer rbenv-help output for printing

TheDen:
This PR:

* Buffers the `help` output so that it is printed all at once, rather than a "scrolled" output
* L103 requires a trailing space after a newline, otherwise bash strips newlines when strings are "psuedo-returned" from a function

Prior (unbuffered)
rbenv-unbuffered

Buffered
rbenv-buffered

### DIFF
--- a/libexec/rbenv-help
+++ b/libexec/rbenv-help
@@ -100,14 +100,16 @@ print_summary() {
   eval "$(documentation_for "$command")"
 
   if [ -n "$summary" ]; then
-    printf "   %-9s   %s\n" "$command" "$summary"
+    printf "   %-9s   %s\n " "$command" "$summary"
   fi
 }
 
 print_summaries() {
+  local buffer=""
   for command; do
-    print_summary "$command"
+    buffer="${buffer}$(print_summary "$command")"
   done
+  printf "${buffer}"
 }
 
 print_help() {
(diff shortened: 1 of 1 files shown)


TheDen:
Hey Mislav 👋, sorry I should have made a case for it first.

If the default _help_ output seems like it's lagging to print for the end user, it lowers the user's confidence in the cli tool's quality. I'd also consider it be an UX issue to have the end user try read text immediately after each block is quickly printed (The GIFs don't capture it well enough unfortunately) for what is less than a paragraph of total text.

I should also note that `rbenv`+`pyenv` is the only cli tool I have installed which has this effect, so it's jarring every time I run into it (enough for me to make this PR). So I'd argue it's "bad" because it's not the normal behaviour of what users have come to expect from cli tools when they print their help output (perhaps even anti-pattern?).

TheDen:
Hey @mislav any update to my previous post? I'll understand if you think this is unimportant, unnecessary etc., I'd just like some closure on this Pr (I'm doing a cleanup of my PRs, and this is over a year old 😮)

TheDen:
Also "scrolled" output wasn't the right term, sorry for the confusion there.

jasonkarns:
Personally, I'm 👎 on this.

While I can appreciate there is a visual difference between progressive rendering and buffered, I disagree with both of these assertions:

> it lowers the user's confidence in the cli tool's quality.

> UX issue to have the end user try read text immediately

If we agree that it takes, let's say 2 seconds to completely render the output; and current behavior is to progressively write lines throughout the duration of that 2 seconds, isn't it a performance _degradation_ to hold all that output in a buffer before it is printed? I'd be more wary of tools that have an initial lag (as would be indicated with a buffer-induced delayed print) because that would typically indicate non-trivial boot-up time. Printing ASAP indicates no such boot-up penalty.

As for reading text immediately, there isn't any rendering of the output that would make the printed text reflow as new output is written. (ie, the jarring reflow of text in an HTML page as image assets are loaded and force reflows). With this output, once the text is printed, it's stationary. So I'd argue that it's a performance and UX _improvement_ to be able to start reading the output as soon as possible. (of course, without any actual UX studies we're just stating our personal opinions; but this is mine 😸 )

More importantly, for *nix tools, buffering is (usually) an anti-pattern because it consumes memory. Being able to pipeline the output from one command to a chain of many others without concern (generally) for the size of the text is a benefit of unbuffered streaming. Granted, the size of the output we're talking about here for rbenv is beyond negligible; and users are unlikely to be piping help output to many other tools. (Other than perhaps less/more/grep) But we'd be behaving counter to other unix tools for little to no reason (IMO).

> I should also note that rbenv+pyenv is the only cli tool I have installed which has this effect, so it's jarring every time I run into it

With this I agree (and add `nodenv` to the list of affected tools), and it's due to how the help text is generated. Rather than having the help output exist within the tool as static content, it is generated dynamically by parsing the comments at the beginning of its own scripts, and by searching within the tool to find the available commands. In a sense, rbenv is "self-documenting". I would be inclined to PRs that improve the _real_ performance of the help command instead of buffer tricks for perceived performance.

TheDen:
Thanks for your thoughts @jasonkarns. I agree, it would be better to put effort in speeding up how the documentation is generated (if possible, there's a lot of string manipulation going on). Typically (from what I've seen) in cli tools the help output isn't generated and simply printed, so imo there is already performance overhead of generating every time `rbenv` is run—but that's the design choice with has benefit of the help pages being self-documenting and modular. Personally I'd generate it once and and have it baked in—a small cost to save end user's CPU cycles + IO.

I think the question here is whether it ought to be printed early or not? Say an error occurs in one of the summary files, it would print partly to stdout then fail, so it wouldn't be an atomic operation.

PR #1151, opened 2019-03-05
shims: handle exported SHELLOPTS

blueyed:
When SHELLOPTS is exported/set when bash starts up, it will use settings
from there.
A common use case might be to export/use it for "xtrace", but that
should not cause errors inside of rbenv/pyenv for things like `[ -n
"$FOO" ]`.

Ref: https://github.com/pyenv/pyenv/pull/1273
Ref: https://github.com/pyenv/pyenv/pull/1237

### DIFF
--- a/libexec/rbenv-rehash
+++ b/libexec/rbenv-rehash
@@ -39,11 +39,14 @@ remove_prototype_shim() {
 # hard-linked for every executable and then removed. The linking
 # technique is fast, uses less disk space than unique files, and also
 # serves as a locking mechanism.
+# It protects itself against outer options (nounset, pipefail).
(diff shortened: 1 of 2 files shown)


blueyed:
Using priviledged mode might be best, but requires newer `env`: https://github.com/pyenv/pyenv/pull/1237#issuecomment-469777407.

mislav:
Thank you, but I think that if `SHELLOPTS=nounset:pipefail` are exported, they will affect every single rbenv internal, not just the shim script. So it's likely that many more edits would be necessary to fix rbenv. Instead, I would advise that users not export SHELLOPTS globally.

PR #1155, opened 2019-03-28
Support system-wide Ruby installations

FooBarWidget:
Makes rbenv look for Ruby installations in both ~/.rbenv/versions _as well as_ in `RBENV_SYSTEM_VERSIONS_DIR`, if set. This allows installing Ruby versions in a system-wide manner that's available to all users on the system.

### DIFF
--- a/README.md
+++ b/README.md
@@ -143,6 +143,25 @@ installed:
 Version names to rbenv are simply the names of the directories in
 `~/.rbenv/versions`.
 
+#### System-wide Ruby Installations
+
+Rbenv also supports system-wide Ruby installations. Such installations
+are available to all users on the system because they are not located
+in a home directory. This works through the `RBENV_SYSTEM_VERSIONS_DIR`
+environment variable: if it is set, then rbenv will look there
+*in addition to* looking in `~/.rbenv/versions`.
+
+For example, suppose that `RBENV_SYSTEM_VERSIONS_DIR` is set to
+`/usr/local/lib/rbenv/versions`. You might then have those versions
+installed:
+
+* `/usr/local/lib/rbenv/versions/2.4.0/`
+* `/usr/local/lib/rbenv/versions/2.6.2/`
+
+Note that `~/.rbenv/versions` has priority over `RBENV_SYSTEM_VERSIONS_DIR`.
+If a Ruby installation exists in both `~/.rbenv/versions` and
+`RBENV_SYSTEM_VERSIONS_DIR`, then rbenv will use the one in `~/.rbenv/versions`.
+
 ## Installation
 
 **Compatibility note**: rbenv is _incompatible_ with RVM. Please make
@@ -342,6 +361,30 @@ that directory can also be a symlink to a Ruby version installed
 elsewhere on the filesystem. rbenv doesn't care; it will simply treat
 any entry in the `versions/` directory as a separate Ruby version.
 
+#### Installing System-wide Ruby versions
+
+`rbenv install` always installs to `~/.rbenv/versions/`. If you want to
+install a Ruby version system-wide (i.e. you're going to use
+`RBENV_SYSTEM_VERSIONS_DIR`) then here's how to do it with ruby-build:
+
+~~~ sh
+# if ruby-build is installed as an rbenv plugin:
+$ sudo "$(rbenv root)/plugins/ruby-build" 2.6.2 "$RBENV_SYSTEM_VERSIONS_DIR/2.6.2"
+
+# if ruby-build is installed standalone:
+$ sudo ruby-build 2.6.2 "$RBENV_SYSTEM_VERSIONS_DIR/2.6.2"
+~~~
+
+The use of `sudo` in the above example is under the assumption that
+`RBENV_SYSTEM_VERSIONS_DIR` is only writable by root. You can omit sudo
+if the directory is writable by the current user. Note however that you
+should then carefully think about what appropriate permissions should be:
+it is generally a bad idea for a Ruby installation directory to be writable
+by multiple users.
+
+Again (as an alternative to using ruby-build), you can download and compile
+Ruby manually as a subdirectory of `RBENV_SYSTEM_VERSIONS_DIR`.
+
 #### Installing Ruby gems
 
 Once you've installed some Ruby versions, you'll want to install gems.
@@ -354,7 +397,8 @@ install gems as you normally would:
 $ gem install bundler
 ```
 
-**You don't need sudo** to install gems. Typically, the Ruby versions will be
+Unless the corresponding Ruby installation was installed system-wide,
+**you don't need sudo** to install gems. Typically, the Ruby versions will be
 installed and writeable by your user. No extra privileges are required to
 install gems.
 
@@ -368,7 +412,7 @@ $ gem env home
 ### Uninstalling Ruby versions
 
 As time goes on, Ruby versions you install will accumulate in your
-`~/.rbenv/versions` directory.
+`~/.rbenv/versions` directory (or in `RBENV_SYSTEM_VERSIONS_DIR`).
 
 To remove old Ruby versions, simply `rm -rf` the directory of the
 version you want to remove. You can find the directory of a particular
@@ -376,7 +420,8 @@ Ruby version with the `rbenv prefix` command, e.g. `rbenv prefix
 1.8.7-p357`.
 
 The [ruby-build][] plugin provides an `rbenv uninstall` command to
-automate the removal process.
+automate the removal process. That command however only supports
+removing from `~/.rbenv/versions`, not from `RBENV_SYSTEM_VERSIONS_DIR`.
 
 ### Uninstalling rbenv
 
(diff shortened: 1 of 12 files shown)


jasonkarns:
One can already control where rbenv versions are located with `RBENV_ROOT`, is that not sufficient?

FooBarWidget:
It is not. Setting `RBENV_ROOT` to a single directory would mean every user has to have write access to that directory in order to install Rubies. My patch introduces a *secondary* directory in which to *look* for Rubies, but *installing* Rubies would still go to `~/.rbenv`.

mislav:
Hi @FooBarWidget thank you for your contribution!

The code looks solid. I was, however, thinking about how "RBENV_SYSTEM_VERSIONS_DIR" doesn't sit perfectly with me as it's an oddly specific name for something that's basically a secondary RBENV_ROOT.

How would you feel about this:

```sh
export RBENV_ROOT="$HOME/.rbenv:/usr/local/lib/rbenv"
```

Basically, RBENV_ROOT becomes a list of paths to search for installed Ruby versions, global version setting, and plugins.

FooBarWidget:
No objections to that approach in principle, but there is one complication: all plugins, including ruby-build/rbenv-install, will have to be updated to support this new `RBENV_ROOT` format. Are you okay with that?

FooBarWidget:
@mislav What's your opinion on the above?

mislav:
> all plugins, including ruby-build/rbenv-install, will have to be updated to support this new `RBENV_ROOT` format.

That's a good point. 🤔 With that in mind, it might be a better approach to designate a new environment variable.

However, since `libexec/rbenv` is the entrypoint for all rbenv commands (and plugin operations), and since that's also the point where RBENV_ROOT is filled if one doesn't exist yet, theoretically it could take an RBENV_ROOT that has colons in it and transform it to one that only has a single path entry. This might be too magic of a behavior in practice. Let me sit on this for a few days and get back to you.

FooBarWidget:
Let me explain my use case a bit better. I am looking to distribute Ruby binaries in the form of Debian/RPM packages.

The problem with the current distributions' Ruby packages are:

 - They only supply a few Ruby versions, but the Ruby versions actually in use by people is much more diverse. I want to supply packages for many more Ruby versions.
 - They only supply the latest patchlevel version of a minor version, but people often lock down to a specific patchlevel version. I want to supply packages for all patchlevel versions too.
 - They don't allow easy switching between different Ruby versions. In Debian, they do it with binary suffixes, e.g. /usr/bin/ruby2.4 and /usr/bin/ruby2.5. But that's difficult to use an error prone, and not supported by a lot of tooling (e.g. scripts with '#!/usr/bin/env ruby').

    It would be great if the Ruby binary is just called 'ruby', but OS support for this varies. The Debian alternatives system only supports a global setting, but the way people use Ruby shows that being able to have a different Ruby on a per-user or even per-app basis is important.

   The Ruby version switching problem is effectively solved by Ruby version managers such as rbenv, so I want to supply packages that integrates with rbenv.

I'm thinking about supply packages that just drop a bunch of files in /usr/lib/rbenv/versions/<version>.

FooBarWidget:
Hi @mislav, have you been able to come to a conclusion?

FooBarWidget:
Hi @mislav, it's been a while, can I bother you to have a look at this again?

mislav:
@FooBarWidget Just dropping a note that I'm taking this into consideration yet again. Doing some code experimentation on my own end first. Please let me know if your requirements changed in the meantim

FooBarWidget:
Thanks mislav. No core requirements changed, but I did receive a bunch of feature requests that may be worth taking into consideration.

- A secondary path (besides `RBENV_SYSTEM_VERSIONS_DIR`) to look in for Ruby installations, so that users can use Fullstaq Ruby packages *as well as* system-wide-installed custom Rubies. So in this scenario, rbenv would look in ~/.rbenv/versions, /usr/local/lib/rbenv/versions, /usr/lib/rbenv/versions. See https://github.com/fullstaq-labs/fullstaq-ruby-server-edition/issues/28
  This suggests that maybe RBENV_SYSTEM_VERSIONS_DIR ought to be a list of paths instead of a single path.
- Secondary paths to look in for plugins. So something like this: primary=~/.rbenv/plugins, secondary=/usr/local/lib/rbenv/plugins:/usr/lib/rbenv/plugins. See https://github.com/fullstaq-labs/fullstaq-rbenv/issues/2
- Support for gem commands located in the user's home directory, instead of only those in the Ruby installation's bin directory. https://github.com/fullstaq-labs/fullstaq-rbenv/pull/1

chhex:
@MaithaM1 @mislav  This is really a very useful proposition  for a i can imagine general requirement as @FooBarWidget  described in his use case.  Basically it is a solution have best of both worlds:   centrally installed and controlled rubies & gems .... and the flexibility of the per user based installation , for which rbenv works so great.... 

Can we expect here anything soon?

EugenMayer:
This PR is open for about 2 years - you can surely consider the effort dead (not happy about it either f

chhex:
@EugenMayer ohh ok, i see, so basically rbenv is "dead" , respectively to be used as it is?.. hmm would have been great this PR. So then i am back to installing packages...   On Centos 7 for example the official ruby rpm is 2.0.0 something ;-/ and sometimes it would be great to be able to use Jruby, with different paths etc

EugenMayer:
Well i think rbend and chruby (both very mildly maintained at all, see https://github.com/postmodern/chruby) are working and kept alive to do so most probably - but nothing else.

I think the popularity of ruby probably decreased in a familiar fashion as those project started to fall asleep. Maybe they are considered feature-complete, used them both for a long time now with some limitations, but the basics are working fine.

So I guess take it as it is or since it's OSS, fork it and merge all the PR's you like :)

chhex:
@EugenMayer Ok, i always knew about ruby and liked it, but only recently picked up more on it, not because of Rails, but as general purpose scripting languages, preferring it way over python, bash, perl, groovy and such... What i also like is the ability of integrate java libraries via jruby .....  seems i am swimming against the current

EugenMayer:
IMHO you are 4 years behind the pulse here, but please emphasis on IMHO.

I used jruby 6 years ago (with a big rails app), replaced with a spring boot app since 4 years. God I miss nothing about rails+jruby and will never ever touch it again.

I used ruby for devops for 8 years, since about 2-3 years replacing all the rakes and thors with golang. Why? Zero dependencies to run on every platform, fat standalone binary - no fat ENV needed to run at all.
Wished i would have written https://github.com/EugenMayer/docker-sync using golang too.

Not sure how much room there is left for ruby or rails. The cli probably falls to golang (or python).

The backend falls to Java/PHP/Golang and the frontend falls to all the Vue/React/Angulars.

So I guess, it's more a niche today and it will only become smaller (assuming my projection).

I should stop here now, since that is far off topic for this lovely PR here, which i wished to have 2 years ago.

chhex:
@EugenMayer that's really interesting input. Is there a possibility to discuss somewhere else? Sorry if i keep on rambling here .... We have never been doing Rails... but since about +10 years Spring and +5 Spring Boot .... and were / are doing everything with Java.... recently the scripting requirement became more prominent, we we adressed first with Groovy.... but it's packaging system is not so great.... that's where Ruby came in.  Golang is neat, yes maybe / probably also an option.... but it's just another compiled language and typed language, which we  really have with Java..... which again is why we turned to Ruby, for where we explicitly don't want Type safeness. Python would have been probably the more "popular" choice .... but sorry those tabs ...

konsolebox:
I somehow understand why it's difficult to merge this implementation. Personally I think it would be better to have a more generic approach which allows multiple root locations that doesn't only allow different locations for versions but also for plugins and hooks as well.

The problem now is that plugins and user hooks expect that `RBENV_ROOT` does not contain multiple paths separated by a colon.  The currently implementation also allows overriding `RBENV_ROOT` and only assign `~/.rbenv` if nothing is assigned to it.

So perhaps it would be a good idea to just add another variable like `RBENV_SYSTEM_ROOT`.  This variable this can have multiple paths separated by a colon set so location of system-installed versions is not opinionated.  Personally I would prefer installing system-packager-installed versions to `/usr/lib/rbenv/versions`, and allow non-system-packager-installed or root-user-installed versions to `/usr/local/lib/rbenv/versions`.

mislav:
@FooBarWidget Thank you for elaborating, spiking out the implementation, and for your patience. This changeset is more code that I am comfortable accepting into the codebase right now, as I'm trying to get rbenv to a point where it's stable with minimal maintenance and few to no feature changes. For that reason, I am going to close this right now and I hope that a solution for your use-case could me made externally to rbenv.

For example, the system could install a global hook at `/usr/lib/rbenv/hooks/rehash/system_rubies.bash`. The hook could iterate over the list of system rubies and ensure that each one is symlinked into `RBENV_ROOT/versions/*`. Then, as any individual user starts interacting with rbenv, they will automatically get access to all system rubies.

This solution is not ideal, as it oversteps the boundaries by writing into the user's home directory, but I think it's still less intrusive and less risky than this code change.

Finally, rbenv gets a bad rap for being slow, and adding more directories and loops to directory lookups in almost every core command is just going to make it slower. Mainly for that reason I'm trying to resist feature changes.

Thank you again for giving me lots to think about! 🙇

PR #1163, opened 2019-05-17
Added `upgrade` command

xtrasimplicity:
Hi there,

I was a little surprised that there wasn't a way to upgrade rbenv using the rbenv command line, so I've added an `upgrade` command which will upgrade both `git` and `brew`-installed versions of `rbenv` and `ruby-build` (if installed).

This PR adds `rbenv upgrade` and adds the `upgrade` command to the useful commands list in `help`:
```
$ rbenv 
rbenv 1.1.2-3-g6a38b2c
Usage: rbenv <command> [<args>]

Some useful rbenv commands are:
   commands    List all available rbenv commands
   local       Set or show the local application-specific Ruby version
   global      Set or show the global Ruby version
   shell       Set or show the shell-specific Ruby version
   install     Install a Ruby version using ruby-build
   uninstall   Uninstall a specific Ruby version
   rehash      Rehash rbenv shims (run this after installing executables)
   version     Show the current Ruby version and its origin
   versions    List installed Ruby versions
   which       Display the full path to an executable
   whence      List all Ruby versions that contain the given executable
   upgrade     Upgrade rbenv (and ruby-build, if installed) to the latest version

See `rbenv help <command>' for information on a specific command.
For full documentation, see: https://github.com/rbenv/rbenv#readme
```

**Note:** This has not been tested with homebrew, as I don't have access to MacOS at the moment. It works on Linux, though!

Thanks!

### DIFF
--- a/libexec/rbenv-help
+++ b/libexec/rbenv-help
@@ -151,7 +151,7 @@ if [ -z "$1" ] || [ "$1" == "rbenv" ]; then
   [ -z "$usage" ] || exit
   echo
   echo "Some useful rbenv commands are:"
-  print_summaries commands local global shell install uninstall rehash version versions which whence
+  print_summaries commands local global shell install uninstall rehash version versions which whence upgrade
   echo
   echo "See \`rbenv help <command>' for information on a specific command."
   echo "For full documentation, see: https://github.com/rbenv/rbenv#readme"
--- a/libexec/rbenv-upgrade
+++ b/libexec/rbenv-upgrade
@@ -22,19 +22,19 @@ is_app_installed() {
 base_path=$(dirname $0)
 
 if is_git_repository "${base_path}"; then
- cd "${base_path}"
+  cd "${base_path}"
  
- echo "Upgrading rbenv..."
- git pull
-
- ruby_build_path="../plugins/ruby-build"
-
- if [ -d "${ruby_build_path}" ]; then
-  cd "${ruby_build_path}"
-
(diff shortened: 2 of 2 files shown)


FooBarWidget:
I was surprised that `upgrade` upgrades rbenv itself, and not a Ruby version. Perhaps it would be a good idea to rename the command so that it's explicit *what* is being upgraded?

xtrasimplicity:
That's a good point. Perhaps `--upgrade` might be a better command?

jasonkarns:
This functionality is already provided by the [rbenv-update](https://github.com/rkh/rbenv-update) plugin. It updates both rbenv and all plugins that are installed via git. (Though it doesn't handle homebrew.)

xtrasimplicity:
@jasonkarns, ahh, I hadn't seen that - thanks! It would be nice to have something built-in, though. :)

DestyNova:
@jasonkarns could you mention that plugin in the docs? I opened a separate PR mentioning how to manually update `ruby-build` but this sounds much more useful for most users.

jasonkarns:
I'd be hesitant to "recommend" a plugin on the main README that's not maintained by the rbenv org. The readme already mentions the existence of plugins and directs to the wiki page for plugins. It's admittedly not super explicit, but anything more than that would imply approval or support of that plugin by the rbenv org, which isn't the case.

@mislav and @hsbt may feel differently, though?

@xtrasimplicity thank you for the time and work on this submission, but I think this still better belongs as a plugin

PR #1177, opened 2019-08-19
Add a closing double quote to rbenv-init for non fish shell on MacOS …

Maartz:
#1176 
Add a closing double quote at the end of the rbenv-init file.

### DIFF
--- a/libexec/rbenv-init
+++ b/libexec/rbenv-init
@@ -149,6 +149,6 @@ cat <<EOS
   *)
     command rbenv "\$command" "\$@";;
   esac
-}
+}"
 EOS
 fi


jasonkarns:
What's the bug this is for? and which shell?

This PR generates invalid shell. For bash:

```
$ ./bin/rbenv init - bash
export PATH="/usr/local/var/rbenv/shims:${PATH}"
export RBENV_SHELL=bash
source '/Users/code/rbenv/rbenv/libexec/../completions/rbenv.bash'
command rbenv rehash 2>/dev/null
rbenv() {
  local command
  command="${1:-}"
  if [ "$#" -gt 0 ]; then
    shift
  fi

  case "$command" in
  rehash|shell|update)
    eval "$(rbenv "sh-$command" "$@")";;
  *)
    command rbenv "$command" "$@";;
  esac
}"
```

I don't see the unmatched opening quote which this PR attempts to fix.

The output on master looks as intended:

```
$ ./bin/rbenv init - bash
export PATH="/usr/local/var/rbenv/shims:${PATH}"
export RBENV_SHELL=bash
source '/Users/code/rbenv/rbenv/libexec/../completions/rbenv.bash'
command rbenv rehash 2>/dev/null
rbenv() {
  local command
  command="${1:-}"
  if [ "$#" -gt 0 ]; then
    shift
  fi

  case "$command" in
  rehash|shell|update)
    eval "$(rbenv "sh-$command" "$@")";;
  *)
    command rbenv "$command" "$@";;
  esac
}
```

mislav:
Most rbenv users, including myself, use a non-fish shell. If the output of rbenv-init for these shells were broken, we would have noticed by now.  I'm not sure what exactly this fixes, but like @jasonkarns points out, looks like it just breaks things by adding an unmatched quote.

Maartz:
Well, that's the problem on my side, and like #1176 issue report it.
If I do not add a " at the end. It just don't work.
I don't know why.

jasonkarns:
@Maartz Can you drop the output of `rbenv init -` in here? (and perhaps also run [rbenv-doctor](https://github.com/rbenv/rbenv-installer#rbenv-doctor)) My only thought is perhaps your install of rbenv got mangled and needs a fresh install? (Possible, if you were editing the rbenv-init file. I'd ensure a fresh checkout of master if installed from git; or update from homebrew)

Issue #1185, opened 2019-09-15
May `version` file ignored if `.ruby-version` file exists under $HOME directory?


ikeyu0806:
I used Ruby 2.5.0 and trying to switch to 2.6.2.
But version did not change even if I use `rbenv glabal`.

Apparently `$HOME/.rbenv/version` was not reflected because this `$HOME/.ruby-version` was loaded.

It took time to identify the cause.So how about outputting a message when `version` file is under the home directory?

jasonkarns:
`rbenv version` already emits the filename of the file (or variable) that is affecting the selected version:

```
$ rbenv version
system (set by /usr/local/var/rbenv/version)

$ echo 2.6.1 > ~/.ruby-version

$ rbenv version
2.6.1 (set by /Users/jasonkarns/.ruby-version)

$ rbenv shell 2.3.3

$ rbenv version
2.3.3 (set by RBENV_VERSION environment variable)
```

ikeyu0806:
It was already Acknowledged.

It seems that the version of rbenv was old.

PR #1186, opened 2019-09-15
Add message to rbenv-global when use .ruby-version

ikeyu0806:
for https://github.com/rbenv/rbenv/issues/1185

I used Ruby 2.5.0 and trying to switch to 2.6.2.
But version did not change even if I use `rbenv glabal`.

Apparently `$HOME/.rbenv/version` was not reflected because There is `$HOME/.ruby-version` the directory I was in.

It took time to identify the cause.So how about outputting a message when `version` file is under the home directory?


ikeyu0806:
It seems that the version of rbenv was old🙏

Thank you for @jasonkarns 
https://github.com/rbenv/rbenv/issues/1185#issuecomment-531576433

PR #1192, opened 2019-10-04
Add attention message about anti virus software

y-vectorfield:
This modification is related to issue [#1334](https://github.com/rbenv/ruby-build/issues/1334) .
If we have installed anti-virus-software on our pc, we should stop it before our `rbenv install`. 
If we use this without stopping, we cannot install ruby ver.xxx. Hence, I added this modification.

### DIFF
--- a/README.md
+++ b/README.md
@@ -193,6 +193,10 @@ If you're on macOS, we recommend installing rbenv with
 5. That's it! Installing rbenv includes ruby-build, so now you're ready to
    [install some other Ruby versions](#installing-ruby-versions) using
    `rbenv install`.
+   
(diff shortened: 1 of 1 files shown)


mislav:
Thank you for letting us know of this, and I'm sorry you had to go through figuring this out before you could install Ruby on your system.

I do not want to add this to our documentation because I don't think it's the responsibility of our documentation to point out shortcomings in some antivirus software for PC. Most people who use rbenv do it so on macOS and Linux and therefore don't ever have to deal with antivirus software.

y-vectorfield:
@mislav 
OK, I understand. Thank you very much for your comments.

PR #1195, opened 2019-10-23
rbenv help: fix 'type: write error: Broken pipe'

FooBarWidget:
Sometimes the command fails with a 'type: write error: Broken pipe'. This is because 'head -1' only reads the first line, then exits. If 'type' writes the second line after 'head -1' has already exited, then the aforementioned error is triggered.

We fix this by buffering the entire output of 'type' before invoking 'head -1'.

### DIFF
--- a/libexec/rbenv-help
+++ b/libexec/rbenv-help
@@ -42,8 +42,13 @@ extract_initial_comment_block() {
 }
 
 collect_documentation() {
+  local all_awks
+  local first_awk
+  all_awks=$(type -p gawk awk)
+  first_awk=$(head -1 <<<"$all_awks")
+
   # shellcheck disable=SC2016
-  $(type -p gawk awk | head -1) '
+  "$first_awk" '
     /^Summary:/ {
       summary = substr($0, 10)
       next


libexec/rbenv-help
```diff
@@ -42,8 +42,13 @@ extract_initial_comment_block() {
...
 
 collect_documentation() {
+  local all_awks
+  local first_awk
+  all_awks=$(type -p gawk awk)
+  first_awk=$(head -1 <<<"$all_awks")
+
   # shellcheck disable=SC2016
-  $(type -p gawk awk | head -1) '
+  "$first_awk" '
```
mislav: Could we try some less verbose approaches? e.g.

```sh
$(type -p gawk awk 2>/dev/null | head -1)
```

FooBarWidget:
It's the bash version that comes with Ubuntu 18.04, 4.4.20. You can see the errors here:
- https://github.com/fullstaq-labs/fullstaq-rbenv/commit/362e522077bbaaa79842796a2d33e2301d552ed4/checks?check_suite_id=277180813
- https://github.com/fullstaq-labs/fullstaq-rbenv/commit/23393dc3f9129b41db0e0e7143850237400cc651/checks?check_suite_id=277171379

libexec/rbenv-help
FooBarWidget: Redirecting all stderr to /dev/null would ignore all non-broken-pipe errors, so I don't think that's a good idea.

I'm not aware of any other way (besides `foo=$(command)`) to fully slurp the output of a command.

If you want shorter, I can change it into this:

~~~bash
  local all_awks
  all_awks=$(type -p gawk awk)

  # shellcheck disable=SC2016
  $(head -1 <<<"$all_awks") '
~~~

But imo that's not clearer.

mislav: > Redirecting all stderr to /dev/null would ignore all non-broken-pipe errors, so I don't think that's a good idea.

That's true, but `type -p` doesn't output any useful errors in case the programs listed weren't found, so I found it acceptable to silence its stderr completely. Thank you!

PR #1203, opened 2020-01-08
libexec/rbenv-version: get rid of misleading "set by $(rbenv-version-origin)" message when system ruby is in use

jf:
as per message: it can be that when the system ruby is in use that there simply is no rbenv version file. However, the output of `rbenv version`, `rbenv versions`, and `rbenv version-name` can be misleading, and point to a missing rbenv version file.

### DIFF
--- a/libexec/rbenv-version
+++ b/libexec/rbenv-version
@@ -8,9 +8,11 @@
 set -e
 [ -n "$RBENV_DEBUG" ] && set -x
 
-RBENV_VERSION_ORIGIN="$(rbenv-version-origin)"
-if [ -z "$RBENV_VERSION_ORIGIN" ]; then
-  echo "$(rbenv-version-name)"
+version_name="$(rbenv-version-name)"
+version_origin="$(rbenv-version-origin)"
+
+if [ "$version_origin" = "${RBENV_ROOT}/version" ] && [ ! -e "$version_origin" ]; then
+  echo "$version_name"
 else
-  echo "$(rbenv-version-name) (set by $RBENV_VERSION_ORIGIN)"
+  echo "$version_name (set by $version_origin)"
 fi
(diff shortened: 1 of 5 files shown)


jf:
What about something like (to give an example):

`system (default; use /home/jf/.rbenv/version to set)`

?

mislav:
@jf Sorry for late reply. Let's just say `system` (nothing in parentheses) for now if `RBENV_ROOT/version` is missing, since I can't think of anything better.

jf:
I did suggest a message; that didn't work for you?

If we just say `system` with no message after, we're back to `rbenv-version-origin` returning nothing, fyi.

mislav:
> I did suggest a message; that didn't work for you?

Yours was a good suggestion; I just think that we shouldn't really say `use /home/jf/.rbenv/version to set` because we don't expect the user to write to this file directly - instead we want them to use `rbenv global <version>`.

Perhaps rbenv should be better at across the board suggesting that a person edits the current version in play by using `rbenv global`, `rbenv local`, and `rbenv shell`. But that's an improvement for another PR.

> If we just say `system` with no message after, we're back to `rbenv-version-origin` returning nothing, fyi.

Well, following my suggestion above, I think that we can leave `rbenv-version-origin` as before your change, and instead implement this in `libexec/rbenv-version`:

```sh
version_name="$(rbenv-version-name)"
version_origin="$(rbenv-version-origin)"

if [ "$version_origin" = "${RBENV_ROOT}/version" ] && [ ! -e "$version_origin" ]; then
  echo "$version_name"
else
  echo "$version_name (set by $version_origin)"
fi
```

jf:
ok, I see (sorry, I got a bit confused there). I'm ok with this: should I push another commit with this then?

mislav:
@jf Yes please! You may even force-push if you want a cleaner history

jf:
:) I prefer the history as is; I'm pushing more commits as we speak.

PR #1220, opened 2020-02-16
Fix fish shell initialization

romanlevin:
Since fish 3.10 at least, the current way of loading rbenv in `fish` is misbehaving, at least in some environments:

```
source: Error encountered while sourcing file '/var/folders/pj/jn249gcn7ddfrjzj2_9mxjhw0000gp/T//.psub.f0iJSWRByB':
source: No such file or directory
```

This changes the initialization to the method recommended by a `fish` developer here: https://github.com/fish-shell/fish-shell/issues/6613#issuecomment-586679958

### DIFF
--- a/libexec/rbenv-init
+++ b/libexec/rbenv-init
@@ -114,7 +114,7 @@ function rbenv
 
   switch "\$command"
   case ${commands[*]}
-    source (rbenv "sh-\$command" \$argv|psub)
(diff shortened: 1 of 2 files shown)


romanlevin:
@mislav Sure, there you go.

romanlevin:
Actually either my current fix is wrong or I'll have to update the tests. Will take a look later.

PR #1241, opened 2020-04-16
Removed duplicate ".ruby-version" mentioning

mikaeldui:
It was listed twice which was confusing to read. I removed the first one as it contained less information.


jasonkarns:
Thanks for contributing! However, those steps aren't duplicates, but are actually distinct. First rbenv looks up the directory tree from the location of the script that is executing. If a .ruby-version file is not found, it does the same lookup beginning from the current working directory.

Longer discussion in the thread that was responsible for the last change to that documentation: https://github.com/rbenv/rbenv/pull/432

PR #1243, opened 2020-04-23
Make work in set -u (nounset) mode

scop:
A couple of low hanging fruits, likely not complete.

### DIFF
--- a/libexec/rbenv-sh-shell
+++ b/libexec/rbenv-sh-shell
@@ -49,7 +49,7 @@ if [ "$version" = "--unset" ]; then
     echo "set -e RBENV_VERSION"
     ;;
   * )
-    echo 'RBENV_VERSION_OLD="$RBENV_VERSION"'
+    echo 'RBENV_VERSION_OLD="${RBENV_VERSION-}"'
     echo "unset RBENV_VERSION"
     ;;
   esac
@@ -107,7 +107,7 @@ if rbenv-prefix "$version" >/dev/null; then
       echo "set -gx RBENV_VERSION \"$version\""
       ;;
     * )
-      echo 'RBENV_VERSION_OLD="$RBENV_VERSION"'
+      echo 'RBENV_VERSION_OLD="${RBENV_VERSION-}"'
       echo "export RBENV_VERSION=\"$version\""
       ;;
     esac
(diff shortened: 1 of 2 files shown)


scop:
I don't actually use rbenv myself at all, nor `set -u` :) But I do use pyenv, and they recommend sending applicable fixes primarily here, so I thought I'd follow that advice. I might start using or experimenting with `set -u` at some point.

So given that, I'm not sure when or if I'll get around do doing more of this.

One way to automate discovery of some issues would be to just `sed -i -e 's/set -e/set -eu/' libexec/*` and make the test suite pass again, but I believe that wouldn't catch at least all of possibly unset variables that end up being used within user's interactive shell (which is what I'm primarily interested in) as opposed to when running those scripts as executables, i.e. in a separate shell (would IMO not be a bad idea to do that either, but it's kind of a separate issue).

Thoughts?

mislav:
> I don't actually use rbenv myself at all, nor `set -u` :) But I do use pyenv, and they recommend sending applicable fixes primarily here, so I thought I'd follow that advice. I might start using or experimenting with `set -u` at some point.
> 
> So given that, I'm not sure when or if I'll get around do doing more of this.

Sure, I understand! But you have to understand as well that I, as a maintainer, prefer to review and merge PRs once they are feature-complete, meaning that they fully solve the problem they intended to solve. Since you described your PR as "likely not complete", it leaves me wondering whether this set of changes is enough to have rbenv/pyenv working on interactive shells that have `set -u`.

> but I believe that wouldn't catch at least all of possibly unset variables that end up being used within user's interactive shell (which is what I'm primarily interested in)

Yeah, the automated approach won't work for this reason. Our tests coverage is primarily over parts of code that always execute in non-interactive bash, and only a few bits of rbenv output are intended to get eval'd in an interactive shell, but we don't have an automated way to verify whether that output would work in a `set -u` shell.

scop:
I tested some more, and the fix seems to be complete -- at least I don't know what else could lurk there or where to look. Changed the syntax a bit from `${foo:-}` to `${foo-}` because it matches the intent more precisely and is easier on the eye.

scop:
I have used the corresponding changes for a long time with pyenv, and have found no issues. Anything more I could do to get this merged? (Please don't ask me to start coding in Ruby, that would be a bit much for a PR this small I think ;))

mislav (approved):
Thank you for the updates!

PR #1244, opened 2020-04-24
Add `--force to Rehash Command

tubbo:
The `rbenv rehash` subcommand can leave the `.rbenv-shim` file around
even though the rehash is clearly complete. Additionally, previous
failed rehashes will cause this file to be left around. To prevent
having to `rm` the file prior to running the rehash command, users who
are confident that rehashing right now won't break anything can append a
`--force` or `-f` flag to ignore and remove this file automatically
before rehashing.


tubbo:
who cares amirite

PR #1245, opened 2020-04-28
Add ZPlug installation instructions

andypalmer:
This PR updates the documentation to include installation via ZPlug plugin.

The plugin at https://gitlab.com/RiverGlide/zsh-rbenv.git is pretty straightforward, and would be a good candidate for bringing under the rbenv organisation.
It uses git submodules to checkout rbenv and ruby-build, and the plugin file itself adds rbenv and rbenv-install to the path as well as calling eval $(rbenv init -)

For ZPlug users, this is a clean, portable, easy way to install

There is also a plugin for rbenv-vars at https://gitlab.com/RiverGlide/zsh-rbenv-vars.git

### DIFF
--- a/README.md
+++ b/README.md
@@ -41,6 +41,8 @@ RVM?**](https://github.com/rbenv/rbenv/wiki/Why-rbenv%3F)
 * [Installation](#installation)
   * [Homebrew on macOS](#homebrew-on-macos)
     * [Upgrading with Homebrew](#upgrading-with-homebrew)
+  * [via ZPlug plugin manager for Zsh](#via-zplug-plugin-manager-for-zsh)
+    * [Upgrading with ZPlug](#upgrading-with-zplug)
   * [Basic GitHub Checkout](#basic-github-checkout)
     * [Upgrading with Git](#upgrading-with-git)
     * [Updating the list of available Ruby versions](#updating-the-list-of-available-ruby-versions)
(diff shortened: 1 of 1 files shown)


mislav:
Hi, I haven't heard of ZPlug, and I don't usually give prominent README placement to projects I have no experience with.

Our regular installation instructions will work for users of zsh as well.

andypalmer:
Hi @mislav, thanks for responding.

If you decide to revisit this in the future, I can raise another PR.
In the meantime, here are some of the reasons that I thought that it was worth including.

[ZPlug](https://github.com/zplug/zplug) is a plugin manager for zsh. There are other shell managers for zsh, such as oh-my-zsh/prezto, but like `RVM`, these managers can be quite invasive under the covers as well as coming packaged with a lot of optional functionality.
ZPlug is closer to the UNIX philosophy of doing one thing well; ZPlug only installs and activates the plugins, it doesn't try to do anything else on top.
ZPlug feels similar to the (sadly, rarely used) installation of UNIX packages to the `/opt` directory, where everything related to one package is stored in the same place.
It also feels very similar to `rbenv` in the way that the plugins are self-contained, much like each version of Ruby has it's own set of gems

## Why use ZPlug to install rbenv?
* ZPlug does not require elevated privileges to install. Both `brew` (for initial installation) and `apt-get`/`yum` etc require super-user access
* Each plugin is installed in isolation. They are responsible for adding their binaries to the path and calling any initialisation functions. This means that the .zshrc is entirely intention revealing. There is no additional noise of extending the PATH variable or calling init functions. ([my current .zshrc](https://gitlab.com/snippets/1975721) - the main noise is configuration for the powerlevel9k prompt) 
* It's portable. The same zshrc works on any operating system (as long as git is installed)
* It's a great compromise between automatic installation via system tools and manual installation via git. You get the benefits of both.
* It's easy to temporarily disable (remove the line from .zshrc and start a new shell) or remove (remove the line and call zplug clean)
* It's easy to upgrade (zplug update)

## Downsides (and mitigations)
* The RiverGlide rbenv plugin relies on git submodules, which means it has to be kept in step with the rbenv release. This is no different to the rbenv brew recipe or OS packages, and could be automated through github actions if the plugin was brought under the rbenv organisation
* It's zsh specific
* It's not the common way to do things (but once upon a time, neither was rbenv or bundler)

Issue #1130, opened 2018-11-28
Instructions in README and `rbenv init` are misleading a bit for Ubuntu


abotalov:
I'm on Ubuntu 18.04 and use Bash.

When I followed step 2 of [Basic Github Checkout in README](https://github.com/rbenv/rbenv/blob/master/README.md#basic-github-checkout), I executed both a step for Bash and for Ubuntu Desktop.

That caused the appearance of `~./bash_profile` which isn't sourced by default in the shell.

So when I executed `rbenv init`, I was advised to append a line to [`~./bash_profile`](https://github.com/rbenv/rbenv/blob/1c772d5ff21bbee52e0c6cdf128c5cffaf8feea8/libexec/rbenv-init#L48) because it existed. That didn't help as the file wasn't sourced.

It looks like quite many users also had such issue - https://github.com/rbenv/rbenv/issues/1067#issuecomment-386901944.

Perhaps, step 2 should clarify that appending a line to `~/.bash_profile` on Ubuntu isn't needed. Also it may be good to improve a detection logic in `rbenv init` to check if `~./bash_profile` is sourced.

mislav:
Thank you for reporting! We should definitely fix this case. Does anyone have ideas how to reliably detect when the current shell is set up to execute `~/.bashrc` but not `~/.bash_profile`? I guess that depends on whether or not bash was started in “login” mode or not.

This is our current logic, and it might not be great: https://github.com/rbenv/rbenv/blob/59785f6762e9325982584cdab1a4c988ed062020/libexec/rbenv-init#L47-L52

maxim-belkin:
`~/.bash_profile` should be sourced by Bash by default, see INVOCATION section in `man bash`. However...

`rbenv init -` does the following (on macOS + Homebrew):

```
export PATH="/Users/mbelkin/.rbenv/shims:${PATH}"
export RBENV_SHELL=bash
source '/usr/local/Cellar/rbenv/1.1.1/libexec/../completions/rbenv.bash'
command rbenv rehash 2>/dev/null
rbenv() {
  local command
  command="$1"
  if [ "$#" -gt 0 ]; then
    shift
  fi

  case "$command" in
  rehash|shell)
    eval "$(rbenv "sh-$command" "$@")";;
  *)
    command rbenv "$command" "$@";;
  esac
}
```
[...]

PR #1254, opened 2020-05-27
Update Readme with instruction for Windows Subsystem for Linux

Pauloparakleto:
Hi @mislav and @jasonkarns, after testing rbenv inside Windows Subsystem for Linux, I decide to give a little contribution to this repo. The second one is gona be on the rbenv-doctor repo. The reason is that there is a condition inside rbenv-doctor file that never reach a .bash_profile.
:)

### DIFF
--- a/README.md
+++ b/README.md
@@ -230,7 +230,7 @@ a systemwide install.
 2. Add `~/.rbenv/bin` to your `$PATH` for access to the `rbenv`
    command-line utility.
 
-   * For **bash**:
+   * For **Bash (and Windows Subsystem for Linux)**:
      ~~~ bash
      $ echo 'export PATH="$HOME/.rbenv/bin:$PATH"' >> ~/.bash_profile
      ~~~


README.md
```diff
@@ -230,7 +230,7 @@ a systemwide install.
 2. Add `~/.rbenv/bin` to your `$PATH` for access to the `rbenv`
    command-line utility.
 
-   * For **bash**:
+   * For **Bash (and Windows Subsystem for Linux)**:
```
jasonkarns, line 233: I'm curious about this callout. WSL functions very much like an OS, right? It's still the Bash shell running on WSL (one could run other shells, like ZSH, on WSL) We don't call out "bash on macOS, Bash on FreeBSD", etc. Is there a particular reason to call out WSL specifically?

hsbt: I have an environment that is fish with WSL2 with my machine. I think this change is miss-reading sentence.

Pauloparakleto: Hi, @jasonkarns and @hsbt, thanks for your feedback, I saw that this issue is related to Ubuntu users. Founding this issue #1130, the question made by @mislav is still relevant. Makes more sence now that Windows 10 might run WLS.
@jasonkarns, about your question, I had forgoten there is not only Ubutun possible to run in WSL :D

ADTC:
FWIW, this works perfectly with the Windows Subsystem for Linux. It is a proper Linux installation (Ubuntu), so technically speaking you're _not_ supporting Windows (the system architecture) by saying that this works in the Windows Subsystem for Linux. I'd want this wording added because it means you're still supporting Windows (as the host environment for the Linux system architecture).

README.md
shanajhwhy: _********************_

PR #1260, opened 2020-08-03
Rename CONDUCT.md to CODE_OF_CONDUCT.md

jasonkarns:
same as: https://github.com/rbenv/ruby-build/pull/1461 and will merge soon as 1461 was approved as well.

> GitHub's expected filename is CODE_OF_CONDUCT.md. Whether that is the best name or even the current de facto standard is to be debated. But moving forward, this is likely the filename that will become the most common within the community, as that is the name GitHub expects and suggests.

> This simply renames the existing code of conduct file from CONDUCT.md to CODE_OF_CONDUCT.md to conform to GitHub's preference.


hsbt:
👍

PR #1279, opened 2020-11-06
README: rearrange rbenv installation sections to find out which section to read more easily

shu-i-chi:
## Changes
* Promoted rbenv-installer as an installation section "Automated Installation with rbenv-installer".
* Rearranged the oreder of the Installation sections (general way, general way, specific OS way):
  1 Homebrew on macOS, 2 Basic GitHub Checkout -> 1 rbenv-installer, 2 Basic GitHub Checkout, 3 Homebrew on macOS

## Reasons of the Changes
I can't tell which installation section to read **at a glance**.

In my opinion, my difficulties comes from;
  * rbenv-installer's link is small, though it is a "way" of rbenv installation same as the rest two (Homebrew & GitHub checkout).
  * The installation section starts with the specific OS (macOS) procedure, and follows the general one.
    * And, there is no word "general" or something in "Basic GitHub Checkout" section's lead description.
      This makes not a mac user (me) to continue to read and find out keywords like "bash" to know that this section is for me.

### DIFF
--- a/README.md
+++ b/README.md
@@ -39,11 +39,12 @@ RVM?**](https://github.com/rbenv/rbenv/wiki/Why-rbenv%3F)
   * [Choosing the Ruby Version](#choosing-the-ruby-version)
   * [Locating the Ruby Installation](#locating-the-ruby-installation)
 * [Installation](#installation)
-  * [Homebrew on macOS](#homebrew-on-macos)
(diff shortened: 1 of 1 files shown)


mislav:
Thank you, but I believe the current documentation is okay. I prefer if people with Homebrew install using brew instead of basic git checkout. And, if someone goes for the git checkout, they will see the link for rbenv-installer easily.

PR #1286, opened 2020-12-05
rehash: use associative array to hold registered shims for a speedup

scop:
Note: I'm not a rbenv user, this is untested. I've used this with pyenv, more details and results before/after for the same change there at https://github.com/pyenv/pyenv/pull/1749/commits/c27937d2c7c995fb14742d82d3576068cd8149c0

### DIFF
--- a/libexec/rbenv-rehash
+++ b/libexec/rbenv-rehash
@@ -101,34 +101,62 @@ make_shims() {
   done
 }
 
-registered_shims=" "
+if ((${BASH_VERSINFO[0]} > 3)); then
 
-# Registers the name of a shim to be generated.
-register_shim() {
-  registered_shims="${registered_shims}${1} "
-}
+  declare -A registered_shims
 
-# Install all the shims registered via `make_shims` or `register_shim` directly.
-install_registered_shims() {
-  local shim file
-  for shim in $registered_shims; do
-    file="${SHIM_PATH}/${shim}"
-    [ -e "$file" ] || cp "$PROTOTYPE_SHIM_PATH" "$file"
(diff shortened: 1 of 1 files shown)


scop:
Guess the CI failures mean bash 3.x is still wanted :I

mislav:
Thanks for offering to speed this up!

Yeah, bash 3.2 support is a must for this project. Please do let us know if you get some optimization working for all versions of bash! Closing this in the meantime

scop:
> Yeah, bash 3.2 support is a must for this project. Please do let us know if you get some optimization working for all versions of bash! Closing this in the meantime since we are not going to maintain two parallel implementations of rehash.

Please note that the PR was already reworked so that it works with all bash versions -- as far as I'm concerned, this could be reopened and applied as-is.

mislav:
@scop I did note that; thank you for your effort!

I did also write this:
> Closing this in the meantime since we are not going to maintain two parallel implementations of rehash.

I didn't merge the PR because it introduces a parallel implementation of rehash based on bash version. In general, I avoid maintaining parallel implementations of anything. Additionally, only one of these implementations is currently exercised in CI. Over time, the other implementation could get broken and we wouldn't notice it in this project until users report the breakage. Therefore, I would rather take a performance penalty and maintain an implementation that works across all supported versions of bash.

Bash 3.2 as a baseline kind of sucks, I know. But it's what we're working with.

scop:
Ok, understandable, just wanted to ensure the rejection was intentional. Luckily for my actual use, pyenv didn't mind and adopted the corresponding change.

PR #1294, opened 2021-01-05
Prevent duplicated shims in PATH variable

felipecrs:
This happens when someone adds the shims folder to the PATH variable manually.

References https://github.com/pyenv/pyenv/pull/1769.

### DIFF
--- a/libexec/rbenv-init
+++ b/libexec/rbenv-init
@@ -84,13 +84,16 @@ fi
 
 mkdir -p "${RBENV_ROOT}/"{shims,versions}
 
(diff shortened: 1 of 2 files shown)


felipecrs:
It was intentionally enabled earlier. I wonder why: if the user has set the PATH variable manually, probably that person doesn't want rbenv to set it again.

lastthyme (approved):

mislav:
Thanks, but see https://github.com/rbenv/rbenv/pull/842#issuecomment-167094835.

PR #1311, opened 2021-03-01
Shell detect improvements

scop:


mislav:
@scop Thank you for the contribution, but this PR had no description, no rationale as to why it's needed, and no tests. Per my maintainership standards, this is not a PR that warrants being merged.

But now that it's in, could you at least tell us more about which cases does this cover, for posterity?

scop:
The commit message covers it, I think:
> For example bash-5.1, bash-static.

mislav:
Thanks! I must admit, I did not look at the commit message.

PR #1325, opened 2021-04-19
Make rbenv path tolerant to brew upgrades

steinybot:
This issue was first discovered in [pyenv](https://github.com/pyenv/pyenv/issues/1716) which copies rbenv.

The problem is that the shims use:
```
exec "/usr/local/Cellar/rbenv/1.1.2/libexec/rbenv" exec "$program" "$@"
```

But this path breaks when rbenv is upgraded and the shims have to be rehashed. This can't be done in the homebrew formula as it runs in a sandbox and it cannot write to `~/.rbenv/shims`.

The fix provided by the [homebrew](https://github.com/Homebrew/homebrew-core/pull/73662#issuecomment-822273295) team is to use `$(brew --prefix)/opt/rbenv/bin/rbenv` instead.


jasonkarns:
Since this is a limitation that only applies to rbenv installed via homebrew (and does not affect other installations of rbenv), I would propose that this is something that should be handled in the [rbenv formula](https://github.com/Homebrew/homebrew-core/blob/master/Formula/rbenv.rb)? I'm thinking of `inreplace` or a `patch` in the formula that replaces `command -v rbenv` with `brew --prefix rbenv` in rbenv-rehash?

PR #1359, opened 2021-11-11
Use test -aP instead of which -a in test helper

scop:
The former is a bash builtin, latter not necessarily available. For
example, `which` is deprecated in Debian's debianutils >= 5.0.


hsbt:
👍

PR #1393, opened 2022-03-08
Simplify version file read

scop:


mislav (approved):
Thank you! Obviously this point of functionality is very critical to rbenv and could break a lot if done improperly, but I can't imagine how this could break and I've tested it thoroughly on bash 3.2 so I would call this a safe improvement from my point of view. 🙇

PR #1414, opened 2022-07-02
Fix bug with bash completion and enabled failglob shell option

e11bits:
I'm using pyenv, which is based on rbenv. The described problem should affect pyenv as well as rbenv, so instead of issuing a pull request for pyenv I'm doing it for rbenv. pyenv states that fixes in rbenv will eventually be applied to pyenv as well. So I hope this is the right thing to do.

Recently, I started using bash with the `failglob` option enabled (`shopt -s failglob`).
With that shell option enabled, I noticed that the completion for pyenv failed with `bash: no match: words[0]`.

The problem seems to be fixed with rewriting the code that removes the first and last array items for completion. This seems also to be more along the line of what is done in rbenv.zsh.

### DIFF
--- a/completions/rbenv.bash
+++ b/completions/rbenv.bash
@@ -5,10 +5,7 @@ _rbenv() {
   if [ "$COMP_CWORD" -eq 1 ]; then
     COMPREPLY=( $(compgen -W "$(rbenv commands)" -- "$word") )
   else
-    local words=("${COMP_WORDS[@]}")
-    unset "words[0]"
-    unset "words[$COMP_CWORD]"
-    local completions=$(rbenv completions "${words[@]}")
+    local completions=$(rbenv completions "${COMP_WORDS[@]: -2:1}")
     COMPREPLY=( $(compgen -W "$completions" -- "$word") )
   fi
 }


completions/rbenv.bash
```diff
@@ -5,10 +5,7 @@ _rbenv() {
   if [ "$COMP_CWORD" -eq 1 ]; then
     COMPREPLY=( $(compgen -W "$(rbenv commands)" -- "$word") )
   else
-    local words=("${COMP_WORDS[@]}")
-    unset "words[0]"
-    unset "words[$COMP_CWORD]"
-    local completions=$(rbenv completions "${words[@]}")
+    local completions=$(rbenv completions "${COMP_WORDS[@]: -2:1}")
```
mislav, line 8: Hi, what the previous logic did was remove the first element of the array and then remove the element at index `$COMP_CWORD`. I do not see COMP_CWORD respected in the new approach. Could you amend the approach so that the old logic is preserved?

mislav (changes requested):
Hi, thanks for attempting to fix this for your shell configuration.

As an alternative approach, would it be an option to add guards to check if the array has any elements before trying to `unset words[0]`? That could perhaps avoid the error that you're seeing.

completions/rbenv.bash
e11bits: Hi, I'm in no way a bash nor a zsh expert, but looking at the zsh implementation it seems to me that the first and last element of the array are removed. 
https://github.com/rbenv/rbenv/blob/c6cc0a1959da3403f524fcbb0fdfb6e08a4d8ae6/completions/rbenv.zsh#L13-L15

And from my understanding, that was also what the bash code was doing. I just used tried to use the bash way for array slices. But again, I could be wrong here.

mislav: bash and zsh have different completion mechanism, so I would not sync logic between them. Instead, the best way to preserve backwards-compatible behavior is to preserve the logic we had before (unless the logic itself was broken)

mislav:
Unfortunately, I had tried to reproduce your problem `shopt -s failglob` turned on in bash, but the completion script just worked. Without a way to confirm the bug, I have no confidence that this change fixes it, or that it's needed at all. Note that pyenv's code is not necessarily up to date with rbenv.

PR #1418, opened 2022-07-15
Fix link to Pow because the server is down

uraitakahito:
The link may have become invalid. This problem may cause a risk of Broken Link Hijacking.


hsbt:
Nice catch. Thanks.

PR #1428, opened 2022-09-18
Simplify resolving symlinks, remove native extension

### LINKED ISSUE #1395: GPL License Compliance Issue
The file shobj-conf in your project is released under the GPL-3.0-or-later license. According to the GPL-3.0 license, the overall project also needs to be released under GPL-3.0-or-later license. The problem is that your project is licensed under the MIT license.

mislav:
From its earliest days, rbenv aggressively resolved symlinks in its codebase. This happened in several areas:
- Top-level `rbenv` command: to determine the absolute path to `<RBENV_INSTALL_PATH>/libexec` so it can be prepended to PATH to enable `rbenv-<command>` style invocation;
- The `rbenv versions` command: to facilitate `--skip-aliases` functionality, which ultimately speeds up rehash;
- The `rbenv hooks` command canonicalized all printed hook paths.

However, since bash doesn't have a built-in way to resolve symlinks, we depended on external utilities: `readlink` or `greadlink`. Since resolving lots of symlinks using external commands is slow, we used to bundle a native bash extension to gain access to [realpath(3)](https://man7.org/linux/man-pages/man3/realpath.3.html) function, which is indeed fast, but adds significant complexity to our codebase and runtime.

This removes all that complexity in favor of canonicalizing directories using `pwd -P`, which I hope is portable across BSD and Linux machines, or not resolving symlinks at all where I do not believe that is necessary, like in the output of `rbenv hooks`.

After this, the `src/configure && make -C src` installation step is now a no-op (since there is no more native extension), but it's kept for backwards compatibility with existing installer scripts.

Ref. #812
Fixes #1395


mislav:
@yyuu Do you see any downsides if this propagates down to derivative version managers such as pyenv?

yyuu:
To be quite honest, I'm not familiar with all the derivatives of rbenv, but mostly rbenv and pyenv only 😅 As per my understanding, this changes the logic to obtain the _real_ path name and that's it. I have applied equivalent changes to my local installation and so far so good. Plus, we'd be able to presume `pwd -P` is very common since it's a part of [POSIX.1-2008](https://pubs.opengroup.org/onlinepubs/9699919799.2008edition/). Anyways we can actually expect `bash` instead of POSIX sh, so it should just work in theory.

At very least for now, I am NOT foreseeing this to disrupt any concrete use cases which I am caring right now 🤔 I will open some report if I will find something later.

mislav:
@yyuu Thanks for chiming in! I also hope this doesn't break anything. I wasn't at all sure why rbenv originally tried to resolve all symlinks so aggressively; after all, for most purposes it shouldn't matter if a path contains a symlink or not.

I'll get this merged soon and cut a new release of rbenv to eliminate the issue of this codebase containing some GPL'd code; I hope you can then propagate the same changes to pyenv too! 🙇

Issue #1382, opened 2022-01-24
Add a warning or error if GEM_HOME is set


benpjohnson:
If `GEM_HOME` is set this seems to break `rbenv` since all gems go into the same directory regardless of the Ruby version used.

Would it be worth adding an explict check for this case [like rvm does](https://github.com/rvm/rvm/blob/35e894c7070eaabc38d91220f6e26d1585a7098a/scripts/functions/notes/warnings/warning_gem_home#L8).

sosedoff:
Just ran into the same problem and it took a while to realize that presence of $GEM_HOME was breaking gem installs.

konsolebox:
You probably can create your own hook script in `/etc/rbenv.d/hooks/exec/` and/or `/etc/rbenv.d/hooks/install/` that checks if GEM_HOME is set but please don't do any of this in rbenv itself.  GEM_HOME is useful for temporarily or permanently relocating installed gems like on a project with different development versions but the same ruby target.  GEM_HOME does well what it's supposed to do but is wrongly used.

Blame the people that sets it globally instead, especially on a system with multiple Ruby installs.  I can think of ways it is wrongly used, like when setting the default install dir to user's gem directory so `--user-install` does not need to be specified when running `gem install` or so `bundler` always installs to the user gem directory.  These all should be altered by the distro that packages the system Ruby instead.

Issue #1232, opened 2020-03-26
cannot find `--user-install`d gem executable


longspell:
I have the following ruby versions installed via rbenv, `2.0.0-p648`, `2.6.3` and `2.6.5` which both `2.6` versions have `rubocop` installed:

Results when using `2.6.3`:
```
$ rbenv versions
  system
  2.0.0-p648
* 2.6.3 (set by /Users/longspell/.ruby-version)
  2.6.5
$ gem info rubocop

*** LOCAL GEMS ***

rubocop (0.80.1, 0.75.1)
    Authors: Bozhidar Batsov, Jonas Arvidsson, Yuji Nakayama
    Homepage: https://github.com/rubocop-hq/rubocop
    License: MIT
    Installed at (0.80.1): /Users/longspell/.gem/ruby/2.6.0
                 (0.75.1): /Users/longspell/.rbenv/versions/2.6.3/lib/ruby/gems/2.6.0

    Automatic Ruby code style checking tool.
$ rubocop --version
0.80.1
```

Results when using `2.6.5`:
```
$ rbenv versions
  system
  2.0.0-p648
  2.6.3
* 2.6.5 (set by /Users/longspell/test/.ruby-version)
$ gem info rubocop
Ignoring byebug-11.1.1 because its extensions are not built. Try: gem pristine byebug --version 11.1.1
Ignoring io-console-0.5.6 because its extensions are not built. Try: gem pristine io-console --version 0.5.6

*** LOCAL GEMS ***

rubocop (0.80.1)
    Authors: Bozhidar Batsov, Jonas Arvidsson, Yuji Nakayama
    Homepage: https://github.com/rubocop-hq/rubocop
    License: MIT
    Installed at: /Users/longspell/.gem/ruby/2.6.0

    Automatic Ruby code style checking tool.
$ rbenv rehash
$ rubocop --version
rbenv: rubocop: command not found

The `rubocop' command exists in these Ruby versions:
  2.6.3
```

Same results for `solargraph`.  Seems to be trying to use the `~/.rbenv/versions` bin instead of the `~/.gems` bin path.  This happens even if I am on `2.6.5` and do `gem install rubocop`.  I've even gone so far as to `gem uninstall rubocop` and remove all versions from `2.6.3`, but upon `gem install rubocop && rbenv rehash` in `2.6.5` I'm still told `rubocop` only exists in `2.6.3`.

mislav:
I'm sorry you're having trouble!

[...]

PR #1436, opened 2022-09-26
Support GEM_HOME, add limited support for user-installed gems

mislav:
The rehash operation will now discover and create shims for executables in additional locations:
- `~/.gem/ruby/<version>/bin/*`
- `$GEM_HOME/bin`

The `rbenv which` (and thus `rbenv exec`) command will also search these locations when looking up a command. This enables shims to dispatch calls to executables added by `GEM_HOME=/my/path gem install` or `gem install --user-install`.

Note that this support is limited:
- It will only work with C Ruby, as it's difficult to guess the `~/.gem/<engine>/<version>` directory for other Rubies without actually loading Ruby;
- It will only work for RBENV_VERSION values in the format `X.Y.Z` and not "system".

Fixes https://github.com/rbenv/rbenv/issues/1232, fixes https://github.com/rbenv/rbenv/issues/1382
Closes #639
Supersedes https://github.com/mislav/rbenv-user-gems

### DIFF
--- a/libexec/rbenv-rehash
+++ b/libexec/rbenv-rehash
@@ -104,6 +104,15 @@ list_executable_names() {
       echo "${file##*/}"
     done
   done
+  # list user-install'ed executables
+  for file in ~/.gem/ruby/*/bin/*; do
+    echo "${file##*/}"
+  done
+  if [ -n "$GEM_HOME" ]; then
+    for file in "$GEM_HOME"/bin/*; do
+      echo "${file##*/}"
+    done
+  fi
 }
 
 # The basename of each argument passed to `make_shims` will be
--- a/libexec/rbenv-which
+++ b/libexec/rbenv-which
@@ -41,6 +41,17 @@ if [ "$RBENV_VERSION" = "system" ]; then
     RBENV_COMMAND_PATH="$(command -v "$RBENV_COMMAND" || true)"
 else
   RBENV_COMMAND_PATH="${RBENV_ROOT}/versions/${RBENV_VERSION}/bin/${RBENV_COMMAND}"
+  if [ ! -x "$RBENV_COMMAND_PATH" ]; then
+    # discover user-install'ed gem executables
+    if [ -n "$GEM_HOME" ]; then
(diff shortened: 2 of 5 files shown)


hsbt:
👍

konsolebox:
Any thoughts on just moving the default user directory to another directory instead, like `~/.rbenv/user`? Using `~/.gem` may conflict with gems installed using `system`.

konsolebox:
With the following code I am able to define a different user directory, and in this case the user directory becomes the default directory itself, but having this new feature that discovers `~/.gem` will make avoiding use of `~/.gem` pointless.

`/etc/rbenv.d/install/custom_defaults/operating_system.rb`:
```
# frozen_string_literal: true

module Gem
  module OperatingSystemDefaults
    def user_dir
      default_dir
    end

    def default_path
      path = []
      path << default_dir
      path << vendor_dir if vendor_dir && File.directory?(vendor_dir)
      path
    end
  end

  self.singleton_class.prepend OperatingSystemDefaults
end
```

`/etc/rbenv.d/install/custom_defaults.bash`:

```
function _after_install_hook {
	local defaults_rb_file defaults_dir_path
	[[ ${PREFIX} && -d ${PREFIX} ]] || return
	defaults_rb_file=$(find "${PREFIX}" -path '*/rubygems/defaults.rb')
	[[ ${defaults_rb_file} == */rubygems/defaults.rb ]] || return
	defaults_dir_path=${defaults_rb_file%.rb}
	mkdir -p "${defaults_dir_path}" || return
	cp "${BASH_SOURCE%.*}/operating_system.rb" "${defaults_dir_path}/"
}

after_install _after_install_hook
```

I hope there's a way to disable the functionality at least if it will be kept.

mislav:
> Any thoughts on just moving the default user directory to another directory instead, like `~/.rbenv/user`? Using `~/.gem` may conflict with gems installed using `system`.

How does the conflict happen in a real-world scenario, can you elaborate a bit more please? Also, feel free to open an issue if this change concerns you.

Issue #1437, opened 2022-09-26
rbenv scripts no longer allowed in another directory besides <bin_dir>/../libexec 


konsolebox:
Specifically in this change: https://github.com/rbenv/rbenv/commit/117a38157537eeb59d73bf8a958363688fdf6383#diff-91b53c00df854a3fb9a4dc150a0a3e37d6c6df79aa36b3731f492b7e23a7ffed

```
-bin_path="$(abs_dirname "$0")"
+rbenv_bin="${BASH_SOURCE:-$0}"
+bin_path="$(cd "${rbenv_bin%/*}"/../libexec && pwd -P)"
```

It no longer resolves the rbenv symbolic link and gets the path to the real rbenv script before getting the directory of it, and hardcodingly forces the libexec scripts to be only placed in `[rbenv_bin_dir]/../libexec`.

Because of this, if the `rbenv` symbolic link is installed by a distro's package manager or script to `/usr/bin/rbenv`, the libexec scripts can also only be installed to `/usr/libexec`, and not in custom locations like `/usr/lib/rbenv/libexec`.  Debian as an example uses this location.  See https://packages.debian.org/sid/all/rbenv/filelist.

I also have an ebuild for Gentoo where I place the libexec scripts to `/usr/libexec/rbenv` instead of `/usr/libexec`.

I suggested a solution in the commit to have `bin/rbenv` installed as a script instead of a symbolic link and then make it `source` the real `rbenv` in the preferred location configured by the package. This will allow the location of the real script to be known through `BASH_SOURCE`.

`bin/rbenv`:
```
#!/usr/bin/env bash
source ../libexec/rbenv # Can be modified by package manager
```

`libexec/rbenv`:
```
bin_path=$(cd "${BASH_SOURCE%/*}" && pwd -P)
```

mislav:
Thanks for reporting!

I better understand the breakage now and I really wish I could resolve BASH_SOURCE to a canonical path in case `bin/rbenv` is a symlink from another location. I could restore the previous behavior, which used `readlink/greadlink` to resolve symlinks, but that adds an overhead of an external call to every rbenv invocation.

[...]

PR #1439, opened 2022-09-27
Switch back to resolving symlinks for `rbenv` executable

mislav:
Considerations:

- `./libexec/rbenv` executable is the entrypoint to the program;

- BASH_SOURCE might be the path to a symlink that has activated `./libexec/rbenv`;

- We must resolve the symlink to learn where rbenv's libexec directory is;

- It's not guaranteed that rbenv commands will always remain directly under their own "libexec" directory, since a package maintainer can change that, e.g. rbenv commands are sometimes placed into `/usr/libexec/rbenv/*`;

- Resolving symlinks might fail and in that case we just assume rbenv project layout.

Fixes https://github.com/rbenv/rbenv/issues/1437 /cc @konsolebox

Reverts parts of https://github.com/rbenv/rbenv/pull/1428, therefore reintroducing ~4ms overhead at runtime to properly detect the location of rbenv scripts. I could not find a simpler way to do this; suggestions welcome!

### DIFF
--- a/libexec/rbenv
+++ b/libexec/rbenv
@@ -42,16 +42,39 @@ export RBENV_DIR
 shopt -s nullglob
 
 rbenv_bin="${BASH_SOURCE:-$0}"
-bin_path="$(cd "${rbenv_bin%/*}"/../libexec && pwd -P)"
+if [ -L "$rbenv_bin" ]; then
+  # resolve rbenv symlink to find out where the actual libexec directory is
+  readlink=$(type -p greadlink readlink 2>/dev/null | head -n1)
+  if [ -n "$readlink" ]; then
+    resolved="$($readlink "$rbenv_bin" 2>/dev/null)"
+    if [[ $resolved == /* ]]; then
+      libexec_dir="${resolved%/*}"
+    else
+      libexec_dir="$(cd "${rbenv_bin%/*}/${resolved%/*}" && pwd)"
+    fi
+  else
+    # no readlink available; assume rbenv project layout
+    libexec_dir="${rbenv_bin%/*}"
+    if [[ $libexec_dir == */* ]]; then
+      libexec_dir="${libexec_dir%/*}/libexec"
+    else
+      libexec_dir="${PWD}/libexec"
+    fi
+  fi
+else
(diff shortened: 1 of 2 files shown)


konsolebox:
> I could not find a simpler way to do this; suggestions welcome!
 
How about the solution I suggested where `bin/rbenv` is changed to a calling script instead?

> `readlink=$(type -p greadlink readlink 2>/dev/null | head -n1)`

Can be changed to `readlink=$(type -P readlink) || readlink=$(type -P greadlink)` avoiding the use of
`head` and a pipe which likely summons another subshell.

> `resolved="$($readlink "$rbenv_bin" 2>/dev/null)"`

`$readlink` here might be worth quoting.

`libexec_dir` and `RBENV_ROOT` might also be worth converting to their canonical paths since they are being compared.

mislav:
Excellent suggestions; thank you!

> How about the solution I suggested where `bin/rbenv` is changed to a calling script instead?

That is also valid and I experimented with changing the `bin/rbenv` that ships with rbenv to be a script instead of a symlink to `libexec/rbenv`. However, for backwards compatibility with other rbenv setups that may be using symlinks already, I thought it would be a nice touch that `libexec/rbenv` goes back to resolving symlinks, and if it has that ability, then I would rather keep `bin/rbenv` as a symlink.

konsolebox:
Master version is working again thank you.  Just a minor thing to consider: If `greadlink` or `readlink` exists as an exported function, `type -p` will print nothing but still return 0 exit status.  `type -P` otoh will only return 0 if executable is found in path.  Is there a reason why `type -p` is preferred?

mislav:
Ah, I didn't know that; thanks for explaining the distinction.

I have been using `type -p` for testing support for executables in all my bash scripts for years, and never have I encountered a bug report with that. I did not even know about `type -P`. I don't feel strongly enough about fixing this right now, so let's see if we can get away with this. If someone exports a function called `readlink` from bash, I hope that they make that function behave like readlink.

mislav:
Ah, only now I understand what you were saying: if someone did export a function called `greadlink`, the result will be an empty string. I will quickly fix this; thanks!

Issue #187, opened 2012-02-02
Fall back to other binary in PATH if gem-installed binary not installed in current ruby


gerrit:
I‘ve got a binary (installed by npm) in `/usr/local/bin/lessc` that I would like to use some of the time. But for some Ruby projects I‘m using the LESS gem which installs a binary in `~/.rbenv/shims/lessc`.

`~/.rbenv/shims` is before `/usr/local/bin` in my `PATH`.

I’d like to be able to use the gem-provided binary when I‘m using a ruby version where it is installed, and otherwise (default case for me) fall back to `/usr/local/bin/lessc`.

Currently, I get the message

`rbenv: lessc: command not found

The `lessc' command exists in these Ruby versions:
  1.9.2-p290`

and the only way to use the npm-provided binary by specifying the whole path. It seems to me that the shim binary should fall back to other binaries that are in the `PATH` already if nothing is installed in the current ruby.

rickard-von-essen:
:+1: for this, I think it should be the default behaviour not requiring a plugin. 
 
> If you have a problem with vagrant or heroku commands, use rbenv whence to find which Ruby versions have installed gems for those projects, uninstall those gems and rbenv rehash.
> 
> Vagrant and Heroku toolbelt have a runtime environment that is supposed to be isolated from the Rubies and gems installed for development. This concept is ruined if you have vagrant or heroku gems installed anywhere, so you'd best uninstall them.

@mislav not very useful if you from time to time do patches on Vagrant and therefore want to have it installed from gem when developing, but not in everyday uses since it's a pain to handle plugins etc. when running as a gem.

kamronbatman:
This problem actually makes rbenv very dangerous. It is routine for people to download repos and run them for their specific needs. If Ruby is not your sole (or primary) tech stack, you can get major problems. Example that happened to me.

Cloned a repo and ran it. It installed the required gems for the required Ruby version. rbenv created shims for them.  In this case, the `chromedriver` binary was installed via `chromedriver-helper`.

Even though I do not normally use that Ruby version or that repo, the shims are global none-the-less. When I ran my e2e tests from another tech stack (webpack/selenium/etc), it ran chromedriver, which gave the `rbenv: command not found` under the hood. Rbenv *did not* send a failure signal and the selenium server thought it was running properly.  This meant my tests were timing out and I could not easily find out why.

[...]

Issue #480, opened 2013-11-02
New `exec` semantics broke `rake cucumber`


markoa:
While smoke-testing the latest rbenv on Semaphore, I came across a failure of a simple Cucumber run in a Rails project:

```
$ bundle exec rake cucumber --trace
** Execute cucumber:ok
/home/user/.rbenv/versions/1.9.3-p448/bin/ruby -S bundle exec cucumber  --profile default
/home/user/.rbenv/versions/1.9.3-p448/bin/ruby: no Ruby script found in input (LoadError)
rake aborted!
Command failed with status (1): [/home/user/.rbenv/versions/1.9.3-p448/bi...]
/home/user/123/vendor/bundle/ruby/1.9.1/gems/rake-10.0.3/lib/rake/file_utils.rb:53:in `block in create_shell_user'
/home/user/123/vendor/bundle/ruby/1.9.1/gems/rake-10.0.3/lib/rake/file_utils.rb:45:in `call'
/home/user/123/vendor/bundle/ruby/1.9.1/gems/rake-10.0.3/lib/rake/file_utils.rb:45:in `sh'
/home/user/123/vendor/bundle/ruby/1.9.1/gems/rake-10.0.3/lib/rake/file_utils_ext.rb:40:in `sh'
/home/user/123/vendor/bundle/ruby/1.9.1/gems/cucumber-1.2.0/lib/cucumber/rake/task.rb:104:in `run'
/home/user/123/vendor/bundle/ruby/1.9.1/gems/cucumber-1.2.0/lib/cucumber/rake/task.rb:193:in `block in define_task'
/home/user/123/vendor/bundle/ruby/1.9.1/gems/bugsnag-1.6.3/lib/bugsnag/rake.rb:27:in `block in bugsnag_define_task'
/home/user/123/vendor/bundle/ruby/1.9.1/gems/rake-10.0.3/lib/rake/task.rb:228:in `call'
/home/user/123/vendor/bundle/ruby/1.9.1/gems/rake-10.0.3/lib/rake/task.rb:228:in `block in execute'
/home/user/123/vendor/bundle/ruby/1.9.1/gems/rake-10.0.3/lib/rake/task.rb:223:in `each'
/home/user/123/vendor/bundle/ruby/1.9.1/gems/rake-10.0.3/lib/rake/task.rb:223:in `execute'
[...]

PR #1446, opened 2022-10-07
Enable fallback to PATH from rbenv shims

mislav:
Previously, when someone activated rbenv shim `foo`, rbenv would error out if `RBENV_ROOT/versions/RBENV_VERSION/bin/foo` did not exist.

This change allows `foo` to be additionally looked up in PATH as fallback. This prevents a case where one of the Ruby versions is shadowing a system-wide command of the same name, preventing the use of the global command across all other Ruby versions.

rbenv used to be strict around this: the lack of a fallback mechanism was to avoid ever falling back to system Ruby for invocations like `bundle` if the current Ruby version did not yet install Bundler. The current approach prevents this scenario by explicitly disallowing fallback for the following executables: ruby, rake, gem, bundle, bundler, irb, rdoc, ri.

Fixes https://github.com/rbenv/rbenv/issues/187
Fixes https://github.com/rbenv/rbenv/issues/865 /cc @jasonkarns 

Additionally, this changeset is a followup to https://github.com/rbenv/rbenv/pull/1436 by bumping up GEM_HOME so it takes precedence over executables found in PATH.

@konsolebox: Since you've previously provided helpful warnings around other potentially risky changes, does this one raise any warning flags for you? Thanks

### DIFF
--- a/libexec/rbenv-exec
+++ b/libexec/rbenv-exec
@@ -1,17 +1,21 @@
 #!/usr/bin/env bash
 #
-# Summary: Run an executable with the selected Ruby version
+# Summary: Run an executable with the current Ruby version
 #
-# Usage: rbenv exec <command> [arg1 arg2...]
+# Usage: rbenv exec <command> [<args>...]
 #
-# Runs an executable by first preparing PATH so that the selected Ruby
-# version's `bin' directory is at the front.
+# Runs an executable by first prepending the current Ruby version's
+# `bin' directory to PATH.
 #
-# For example, if the currently selected Ruby version is 1.9.3-p327:
-#   rbenv exec bundle install
+# For example, this invocation:
 #
-# is equivalent to:
-#   PATH="$RBENV_ROOT/versions/1.9.3-p327/bin:$PATH" bundle install
+#   RBENV_VERSION=3.1.2 rbenv exec bundle install
+#
+# roughly translates to:
+#
+#   PATH="$RBENV_ROOT/versions/3.1.2/bin:$PATH" bundle install
+#
+# See `rbenv help which' for information about the fallback mechanism.
 
 set -e
 [ -n "$RBENV_DEBUG" ] && set -x
@@ -41,7 +45,7 @@ for script in "${scripts[@]}"; do
 done
 
 shift 1
-if [ "$RBENV_VERSION" != "system" ]; then
+if [ "$RBENV_BIN_PATH" = "${RBENV_ROOT}/versions/${RBENV_VERSION}/bin" ]; then
   export PATH="${RBENV_BIN_PATH}:${PATH}"
 fi
 exec -a "$RBENV_COMMAND" "$RBENV_COMMAND_PATH" "$@"
--- a/libexec/rbenv-which
+++ b/libexec/rbenv-which
@@ -6,6 +6,17 @@
 #
 # Displays the full path to the executable that rbenv will invoke when
 # you run the given command.
+#
+# If GEM_HOME is set, the result will be `$GEM_HOME/bin/<command>' if
+# it exists.
+#
+# If the current Ruby version is "system", the command will be looked up
+# in PATH.
+#
+# For other Ruby versions, the command is first looked up in
+# `$RBENV_ROOT/versions/$RBENV_VERSION/bin'. If it does not exist and if
+# the command is not one of the core Ruby executables (e.g. "ruby",
+# "gem", "bundle"), it will be looked up in PATH.
 
 set -e
 [ -n "$RBENV_DEBUG" ] && set -x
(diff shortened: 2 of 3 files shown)


konsolebox:
> @konsolebox: Since you've previously provided helpful warnings around other potentially risky changes, does this one raise any warning flags for you? Thanks

I'm quite ok with the fallback code so far.  I've been thinking about the GEM_HOME support `rbenv` will generously provide but ended up having no strong opinion about it, partially because it won't really affect me; another is because `GEM_HOME` will always affect the ruby instance anyway regardless if executables in `${GEM_HOME}/bin` are executed or not.  I only worry about the shim noise it will allocate but it's nothing really.

Is there a reason why `${GEM_HOME}/bin` is not auto-included in PATH?  Not saying it should be.

I also ideally think "version bin path" should be made sure to be only added once, and also only added exactly before the shims path to avoid getting a higher PATH priority from a directory a user may not want it to.

mislav:
> Is there a reason why `${GEM_HOME}/bin` is not auto-included in PATH? Not saying it should be.

You mean, to include it in PATH when `libexec/rbenv` is preparing the final PATH before dispatching other `rbenv-*` commands? We could do that in theory, but I'm more comfortable of scoping any GEM_HOME behavior to just the shim dispatch mechanism.

> I also ideally think "version bin path" should be made sure to be only added once, and also only added exactly before the shims path to avoid getting a higher PATH priority from a directory a user may not want it to.

Sorry, could you elaborate? I'm not sure I follow

konsolebox:
> > Is there a reason why `${GEM_HOME}/bin` is not auto-included in PATH? Not saying it should be.
> 
> You mean, to include it in PATH when `libexec/rbenv` is preparing the final PATH before dispatching other `rbenv-*` commands? We could do that in theory, but I'm more comfortable of scoping any GEM_HOME behavior to just the shim dispatch mechanism. If `${GEM_HOME}/bin` wasn't in PATH already, I'm reluctant to automatically push it into PATH to avoid negatively affecting all rbenv commands and subprocesses (including `ruby` itself).

To be exact it's these lines:

```
if [ "$RBENV_BIN_PATH" = "${RBENV_ROOT}/versions/${RBENV_VERSION}/bin" ]; then
  export PATH="${RBENV_BIN_PATH}:${PATH}"
fi
```
It was originally:
```
if [ "$RBENV_VERSION" != "system" ]; then
...
```
Which seems to make sure only `"${RBENV_ROOT}/versions/${RBENV_VERSION}/bin"` is added to PATH and not `${GEM_HOME}/bin` when `RBENV_COMMAND_PATH` points to an executable in it.

I have no definite opinion about it as well.  Perhaps not adding it to PATH is safer, and maybe more consistent to instances where GEM_HOME is modified somewhere during runtime.

> > I also ideally think "version bin path" should be made sure to be only added once, and also only added exactly before the shims path to avoid getting a higher PATH priority from a directory a user may not want it to.
> 
> Sorry, could you elaborate? I'm not sure I follow

If let's say I have a PATH with a path that comes before `${RBENV_ROOT}/shims` like the `${GEM_HOME}/bin` path which I explicitly add:

```
/home/user/project/common-gems/bin:/home/user/.rbenv/shims:/home/user/.rbenv/bin:/home/user/bin:/home/user/.local/bin:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin:/opt/bin:
```
where `GEM_HOME` is `/home/user/project/common-gems/bin`:

I would prefer `${RBENV_ROOT}/versions/${RBENV_VERSION}/bin` to only be added right before `/home/user/.rbenv/shims` and not at the very beginning of `PATH`.  This is for the sake of consistency of runtime priorities.

Also if `${RBENV_ROOT}/versions/${RBENV_VERSION}/bin` is already in `PATH`, I want that it doesn't add itself again.  This is something that may happen on a recursive call to `rbenv` or one of its `commands`.

mislav:
> To be exact it's these lines:

Ah, I see! I can consider allowing `${GEM_HOME}/bin` to be prepended here; that would make sense. The extra conditional you noticed me adding is because I do not want to prepend the path for a system fallback. For example, if executing `foo` was ultimately found in `/usr/bin/foo`, I do not want `/usr/bin` to be prepended to PATH.

konsolebox:
> Ah, I see! I can consider allowing `${GEM_HOME}/bin` to be prepended here; that would make sense. The extra conditional you noticed me adding is because I do not want to prepend the path for a system fallback. For example, if executing `foo` was ultimately found in `/usr/bin/foo`, I do not want `/usr/bin` to be prepended to PATH.

I'm starting to think that `${RBENV_ROOT}/versions/${RBENV_VERSION}/bin` should not be added to PATH at all since if theoretically a grand child process (likely a shell script) changes directory to a project directory with a `.ruby-version` that specifies a different Ruby version, that version can be ignored since the executables in the previous `${RBENV_ROOT}/versions/${RBENV_VERSION}/bin` will be called first.

Was optimization the intention for adding it?

I also think `${GEM_HOME}/bin` shouldn't also be added as well since some configuration may change its value in the project as well, which could be done in the grand child process shell itself, or some hooks for `rbenv` which alters the variables like this one:

`/etc/rbenv.d/exec/environment.bash`:
```
{
	__=${PWD}

	while [[ $__ == /* ]]; do
		if [[ -e $__/.ruby-environment ]]; then
			[[ -r $__/.ruby-environment ]] && source "$__/.ruby-environment"
			break
		fi

		__=${__%/*}
	done

	unset -v __
}
```
`.ruby-environment`:
```
export GEM_HOME=${BASH_SOURCE%/*}/../common-project-gems
export PATH=${GEM_HOME}/bin:${PATH}
```
The GEM_HOME value example is mostly useful for examining the gems a project pulls in because they're placed in an isolated directory.  They're easy to grep, clean up and refresh.

konsolebox:
If it's for the sake of having newly installed executables detected immediately at the same runtime, the paths should be added after the shims path instead.

For example, `/home/user/.rbenv/shims:/home/user/.rbenv/bin:/sbin:/bin` becomes `/home/user/.rbenv/shims:/home/user/.rbenv/versions/3.0.4/bin:/home/user/.rbenv/bin:/sbin:/bin` instead.

mislav:
> I'm starting to think that `${RBENV_ROOT}/versions/${RBENV_VERSION}/bin` should not be added to PATH at all

I wanted to get away with that as well, but unfortunately, prepending the bin directory to PATH is what's needed to support `ruby -S`: https://github.com/rbenv/rbenv/issues/480

If shims remain in PATH, shelling out to `ruby -S <command>` would fail if command was found among rbenv shims, since shims are bash scripts and not ruby scripts.

`ruby -S` is itself a huge hack and I don't like that we have to work around this but I wasn't able to find other ways around this.

HACKED-need-HELP (changes requested):
Don't know what I'm doing thanks

Jenibabi1984 (approved):

saltman007web (approved):

PR #1450, opened 2022-10-07
Use `read -a` to split output into an array

mislav:
This avoids shellcheck warnings about using the `(...)` array notation to split captured output into an array.

### DIFF
--- a/libexec/rbenv-init
+++ b/libexec/rbenv-init
@@ -126,7 +126,7 @@ if [ -z "$no_rehash" ]; then
   echo 'command rbenv rehash 2>/dev/null'
 fi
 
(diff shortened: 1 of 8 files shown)


libexec/rbenv-init
```diff
@@ -126,7 +126,8 @@ if [ -z "$no_rehash" ]; then
   echo 'command rbenv rehash 2>/dev/null'
 fi
 
-commands=(`rbenv-commands --sh`)
+read -d '' -r -a commands <<<"$(rbenv-commands --sh)" || true
```
konsolebox, line 129: Might be worth specifying `IFS=$'\n'` here as well so it doesn't worry about rogue IFS values.

PR #1461, opened 2022-11-22
Update actions/checkout version to v3

ydah:
The version 3.x appears to be due to the following:
- https://github.com/actions/checkout/pull/689

There have been no breaking changes, and we seem to have no problem keeping up with the latest version.


mislav (approved):

mislav:
Thank you!

PR #1462, opened 2022-11-23
Fix a formatting in ci.yml

ydah:
This PR is fix a formatting in ci.yml.
- trailing spaces
- indentation


mislav (commented):
Thank you, but the current yaml file isn't really broken, so there's nothing to fix. Trailing spaces and the decision to indent YAML lists are opinionated and it doesn't hurt to have a yaml file formatted either way.

PR #1479, opened 2023-02-08
Rt shell test refactor 2

richiethomas:


richiethomas:
@mislav a 2nd set of eyes on these shellcheck errors would be helpful.  Related to [this error](https://www.shellcheck.net/wiki/SC2030) and [this one](https://www.shellcheck.net/wiki/SC2031).

They're both warning that an env var modification was made in a subshell and therefore may be lost in the parent shell. 
 But the specs I'm adding specifically assert that the modifications are visible in the parent shell, so these errors strike me as false positives.

richiethomas:
Closing this since it appears to not be helpful.

PR #1489, opened 2023-03-21
Add rbenv.bats spec to cover the case when RBENV_DIR does not contain a forward slash

richiethomas:
While comparing the specs in `rbenv.bats` with the code in `libexec/rbenv`, I noticed that [this spec](https://github.com/rbenv/rbenv/blob/master/test/rbenv.bats#L34) appears to cover the first half of the `||` condition, when `RBENV_DIR` does contain the `/` character:

```
[[ $RBENV_DIR == /* ]] ...
```

But there doesn't appear to be an equivalent test for the 2nd half of that `||` condition:

```
... || RBENV_DIR="$PWD/$RBENV_DIR"
```

This PR adds a test to cover that code path.

### DIFF
--- a/test/rbenv.bats
+++ b/test/rbenv.bats
@@ -38,6 +38,11 @@ load test_helper
   assert_output "$dir"
 }
 
+@test "inherited RBENV_DIR without a '/' char" {
+  RBENV_DIR="." run rbenv echo RBENV_DIR
+  assert_output "$(pwd)"
+}
+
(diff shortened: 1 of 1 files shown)


richiethomas:
Closing this for now, because I wasn't able to get a "red" result for the test by commenting-out the `||` clause, as part of the "red-green-refactor" pattern.  Will re-open if and when I can solve that.

PR #1497, opened 2023-03-31
ci: trigger differential-shellcheck workflow on push

jamacku:
Fixes: https://github.com/redhat-plumbers-in-action/differential-shellcheck/issues/215

This should get rid of annoying messages from the `github-code-scanning` bot.

Screenshot from 2023-03-22 07-00-32

Also, I have added the required permission for uploading SARIF reports to GitHub

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -2,6 +2,8 @@
 
 name: Lint
 on:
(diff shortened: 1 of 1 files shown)


mislav (approved):
Thank you!

PR #1512, opened 2023-07-02
version-file docs update

richiethomas:
The job of the `rbenv-version-file` command is to print the filepath for the file which sets the current Ruby version.  It searches for a `.ruby-version` file, traversing upward if such a file is not found until it reaches the root directory.

 - If a directory is passed as an argument, the command uses that directory as a starting point.
 - If no directory is passed as an argument:
   - The command first searches in `RBENV_DIR`.
   - If no `.ruby-version` file is found, it searches again in the current directory (provided that `RBENV_DIR` is different from the current directory).
   - If both of those attempts fail to find a `.ruby-version` file, it prints the path to RBENV's global version file (`"${RBENV_ROOT}/version"`).

This global version file may not actually exist, for example in the following cases:

 - If a user has not yet set a global version via `rbenv global <version_number>`.
 - If they have not set a local project version, either via `rbenv local <version_number>` or by manually creating a `.ruby-version` file.

In the case where the filepath returned by `rbenv-version-file` doesn't actually exist, the result might be confusion on the user's part.  This PR attempts to address that confusion, by updating the output of `rbenv help version-file` to specify that this command's output is the filepath which *would* set the Ruby version, *if* it exists.

### DIFF
--- a/libexec/rbenv-version-file
+++ b/libexec/rbenv-version-file
@@ -1,6 +1,9 @@
 #!/usr/bin/env bash
 # Usage: rbenv version-file [<dir>]
-# Summary: Detect the file that sets the current rbenv version
+# Summary: Detect the file that sets the current rbenv version.
+#
+# Prints locations of existing files, but also prints a location
+# of a file that, when created, would have dictated the version.
 set -e
 [ -n "$RBENV_DEBUG" ] && set -x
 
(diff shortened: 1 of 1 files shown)


libexec/rbenv-version-file
```diff
@@ -1,6 +1,10 @@
 #!/usr/bin/env bash
 # Usage: rbenv version-file [<dir>]
-# Summary: Detect the file that sets the current rbenv version
+# Summary: Detect the file that sets the current rbenv version.
```
mislav: Minor: the summaries of other commands don't end in periods, so this one shouldn't either.

mislav (approved):
Thank you! I have pushed amendments to the text for more clarity.

Issue #119, opened 2011-10-11
zsh command hash table is stale with system ruby


bricestacey:
I have a zsh script to setup dev environments that create tmux sessions and windows. (see http://pastie.org/2643820)

Occasionally, a tmux window's zsh's command hash would still be pointing to the system ruby.

```
[master][~/code/ezpz] hash | grep ruby=
jruby=/Users/brice/.rbenv/shims/jruby
ruby=/usr/bin/ruby

[master][~/code/ezpz] env | grep PATH
PATH=/Users/brice/.rbenv/shims:/Users/brice/.rbenv/bin:/usr/bin:/bin:/usr/sbin:/sbin:/usr/local/bin:/usr/X11/bin:/Users/brice/.rbenv/shims:/Users/brice/.rbenv/bin:/opt/local/bin:/usr/local/git/bin

[master][~/code/ezpz] ruby -v
ruby 1.8.7 (2010-01-10 patchlevel 249) [universal-darwin11.0]

[master][~/code/ezpz] which ruby
/usr/bin/ruby

[master][~/code/ezpz] /Users/brice/.rbenv/shims/ruby -v
ruby 1.9.2p290 (2011-07-09 revision 32553) [x86_64-darwin11.0.0]
```

Would it be a good idea to maybe include a call the builtin rehash after rbenv-init has been run? This seems to have solved my problem. I just do the following in .zshrc

```
export PATH="$HOME/.rbenv/bin:$PATH"
eval "$(rbenv init -)"
builtin rehash
```

PR #1534, opened 2023-10-19
Clear command history after 'rbenv rehash' in Fish shell

richiethomas:
I'm reading the code of `libexec/rbenv-sh-rehash`, and I found [this line of code](https://github.com/rbenv/rbenv/blob/c4395e58201966d9f90c12bd6b7342e389e7a4cb/libexec/rbenv-sh-rehash#L18), which says that if the user's shell is Fish, we execute a no-op.  This is different from [what happens with other shells](https://github.com/rbenv/rbenv/blob/c4395e58201966d9f90c12bd6b7342e389e7a4cb/libexec/rbenv-sh-rehash#L21), where we clear the user's command history cache via the "hash -r" command.

It seems like the original call to "hash -r" was added in response to [this Github issue](https://github.com/rbenv/rbenv/issues/119), and that the no-op for the Fish shell was added as part of [this PR](https://github.com/rbenv/rbenv/commit/5bfec84432e63c46df4724918080abe488ba64b4#diff-5e6f35f443d9791b1df6bfa471777ba127d986dbe696840066ff3b8aee8be601).  But I don't see a Github issue related to the latter PR, or any discussion around why the Fish branch is a no-op.  So I'm curious why if there's a reason not to clear the command history if the user's shell is Fish, when it is cleared for other shells.

According to [the Fish docs](https://fishshell.com/docs/current/cmds/history.html), Fish supports the "history" command, which does something similar to Bash's "hash" command.  And "history clear" would achieve the same outcome as "hash -r".  So theoretically we could replace the no-op with "history clear".  According to [the docs](https://fishshell.com/docs/current/cmds/history.html), prefacing the command with "builtin" is the way to circumvent the "Are you sure?" prompt which otherwise appears after running this command.

Screenshot 2023-10-19 at 11 28 47 AM

I believe the `2>/dev/null` and `; or true` code is correct Fish syntax, but could use an extra set of eyes on that to confirm.

### DIFF
--- a/libexec/rbenv-sh-rehash
+++ b/libexec/rbenv-sh-rehash
@@ -15,7 +15,7 @@ rbenv-rehash
 
 case "$shell" in
 fish )
-  echo "builtin history clear"
+  echo "builtin history clear 2>/dev/null; or true"
   ;;
 * )
   echo "hash -r 2>/dev/null || true"
--- a/libexec/rbenv-sh-rehash
+++ b/libexec/rbenv-sh-rehash
@@ -15,7 +15,7 @@ rbenv-rehash
 
 case "$shell" in
 fish )
-  # no rehash support
+  echo "builtin history clear"
   ;;
 * )
   echo "hash -r 2>/dev/null || true"
--- a/test/rehash.bats
+++ b/test/rehash.bats
@@ -158,6 +158,6 @@ SH
 @test "sh-rehash in fish" {
   create_executable "2.0" "ruby"
   RBENV_SHELL=fish run rbenv-sh-rehash
-  assert_success "builtin history clear"
+  assert_success "builtin history clear 2>/dev/null; or true"
   assert [ -x "${RBENV_ROOT}/shims/ruby" ]
 }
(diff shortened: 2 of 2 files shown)


mislav:
Thanks, but you might be misunderstanding what we are clearing in other shells. We are not clearing command history per se, but a cached resolution of a command to its absolute location on disk.

> This is different from [what happens with other shells](https://github.com/rbenv/rbenv/blob/c4395e58201966d9f90c12bd6b7342e389e7a4cb/libexec/rbenv-sh-rehash#L21), where we clear the user's command history cache via the "hash -r" command.

In other shells, what we are actually clearing is the sort of shell session cache that remembers that a previously run `ruby` command resolved to e.g. `/usr/bin/ruby`. After doing the `rbenv rehash` process, there now might be additional entries in PATH (e.g. `ruby` should not resolve to `~/.rbenv/shims/ruby`), and it's good to do `hash -r` to bust the cache.

In fish shell, this cache-busting didn't seem necessary, and so it is a no-op in fish. Did you find evidence that cache clearing is necessary in your case?

richiethomas:
> We are not clearing command history per se, but a cached resolution of a command to its absolute location on disk.

Yep, I deduced that.  I think my mistake was in assuming that Fish's `history` command served the purpose of caching the commands and their locations as well.  But upon further inspection, it looks like Fish may be maintaining that history for other reasons, unrelated to caching the location for faster execution.  This implies that my solution doesn't do the job I intended it to do, after all.

Closing this PR.

PR #1537, opened 2023-11-20
Update alerts in README

dipree:
The old alert syntax was deprecated:

```
> **Note** 
> Old
```

```
> [!NOTE]
> New
```


mislav (approved):
Thank you! TIL

PR #1541, opened 2023-11-30
Move gem-rehash monkey patch

### LINKED ISSUE #1540: `gem-rehash` inop for `bundle install`
It appears that installing executable gems via `bundle install` does not trigger automatic rehashing as would be expected. This can be quickly verified by deleting the shim for an executable, uninstalling its gem and reinstalling it via bundler (the shim does not regenerate).

With some quick and dirty debugging `puts` calls, I've gathered that ~~for some reason (still unknown),~~ the monkey patched method below is never invoked when running `bundle install`

https://github.com/rbenv/rbenv/blob/d10388a0529cb4f05b170395bb01c257a8186062/rbenv.d/exec/gem-rehash/rubygems_plugin.rb#L13-L33

Edit: I've tracked down the reason. It appears that when invoked via `bundler install`, the `rubygem_plugin.rb` load happens _within_ `Bundler::Installer.install` ([here](https://github.com/rubygems/rubygems/blob/9766fa6ba4619571feb13fd30e199c62bcc784f9/bundler/lib/bundler/installer.rb#L23) -> [here](https://github.com/rubygems/rubygems/blob/9766fa6ba4619571feb13fd30e199c62bcc784f9/bundler/lib/bundler/installer.rb#L84) -> [here](https://github.com/rubygems/rubygems/blob/9766fa6ba4619571feb13fd30e199c62bcc784f9/bundler/lib/bundler/installer.rb#L217)). So it's trying to monkey patch the method while it's being run (thus the patched version is never called).

---

debugging `puts` calls:
```ruby
if defined?(Bundler::Installer) && Bundler::Installer.respond_to?(:install) && !Bundler::Installer.respond_to?(:install_without_rbenv_rehash)
  puts 'LOADING BUNDLER PLUGIN'
  Bundler::Installer.class_eval do
    class << self
      puts 'PATCHING INSTALL METHOD'
      alias install_without_rbenv_rehash install
      def install(root, definition, options = {})
        puts 'RUNNING PATCHED INSTALL METHOD' # <= this doesn't get printed
```

benmelz:
The `rubygems_plugin.rb` file doesn't get loaded until after `Bundler::Installer.install` is called, so target `Bundler::Installer#install` to fix invocation order issues.

Closes #1540

### DIFF
--- a/rbenv.d/exec/gem-rehash/rubygems_plugin.rb
+++ b/rbenv.d/exec/gem-rehash/rubygems_plugin.rb
@@ -10,27 +10,27 @@ hook = lambda do |installer|
   end
 end
 
-if defined?(Bundler::Installer) && Bundler::Installer.respond_to?(:install) && !Bundler::Installer.respond_to?(:install_without_rbenv_rehash)
+if defined?(Bundler::Installer) && Bundler::Installer.methods.include?(:install) && !Bundler::Installer.methods.include?(:install_without_rbenv_rehash)
   Bundler::Installer.class_eval do
-    class << self
-      alias install_without_rbenv_rehash install
-      def install(root, definition, options = {})
-        begin
-          if Gem.default_path.include?(Bundler.bundle_path.to_s)
-            bin_dir = Gem.bindir(Bundler.bundle_path.to_s)
-            bins_before = File.exist?(bin_dir) ? Dir.entries(bin_dir).size : 2
-          end
-        rescue
-          warn "rbenv: error in Bundler post-install hook (#{$!.class.name}: #{$!.message})"
-        end
+    private
 
-        result = install_without_rbenv_rehash(root, definition, options)
-
-        if bin_dir && File.exist?(bin_dir) && Dir.entries(bin_dir).size > bins_before
-          `rbenv rehash`
+    alias install_without_rbenv_rehash install
+    def install(options)
+      begin
+        if Gem.default_path.include?(Bundler.bundle_path.to_s)
+          bin_dir = Gem.bindir(Bundler.bundle_path.to_s)
+          bins_before = File.exist?(bin_dir) ? Dir.entries(bin_dir).size : 2
         end
-        result
+      rescue
+        warn "rbenv: error in Bundler post-install hook (#{$!.class.name}: #{$!.message})"
       end
+
+      result = install_without_rbenv_rehash(options)
+
+      if bin_dir && File.exist?(bin_dir) && Dir.entries(bin_dir).size > bins_before
+        `rbenv rehash`
+      end
+      result
     end
   end
 else
--- a/rbenv.d/exec/gem-rehash/rubygems_plugin.rb
+++ b/rbenv.d/exec/gem-rehash/rubygems_plugin.rb
@@ -10,7 +10,9 @@ hook = lambda do |installer|
   end
 end
 
-if defined?(Bundler::Installer) && Bundler::Installer.methods.include?(:install) && !Bundler::Installer.methods.include?(:install_without_rbenv_rehash)
+if defined?(Bundler::Installer) &&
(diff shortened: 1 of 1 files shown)


mislav (commented):
Thanks for tracking this down and for proposing a solution!

`rbenv rehash` should be started concurrently by multiple threads at the same time. Depending on the `bundle install --jobs` setting, multiple gems that each install their executables could be installed at the same time. Could you test to see if the `install` instance method is being invoked per each gem and could potentially be invoked concurrently? In such case, there should be a mutex around `rbenv rehash`.

benmelz:
> Could you test to see if the install instance method is being invoked per each gem and could potentially be invoked concurrently? In such case, there should be a mutex around rbenv rehash.

The parallelization [happens within `Bundler::Installer#install`](https://github.com/rubygems/rubygems/blob/9766fa6ba4619571feb13fd30e199c62bcc784f9/bundler/lib/bundler/installer.rb#L201), so the wrapper should still be run once, sequentially after everything has been spawned and waited on in terms of parallelization.

> Alternatively, would it be an option to hook into Plugin::Events::GEM_AFTER_INSTALL_ALL? I think that such an approach could last longer than any form of monkey patching.

So I actually initially took this approach on [this branch which I just pushed up](https://github.com/benmelz/rbenv/tree/gem-rehash-bundler-plugin). I did manage to get it mostly working, but stopped due to some asterisks:
- The bundler plugin API is heavily tied to the concept that gems provide plugins and plugins provide hooks (like `after-install-all`) - even if you try to programmatically add your own hooks, bundler [will refuse to run them unless there's an associated plugin](https://github.com/rubygems/rubygems/blob/46585225c34676d605daad05604a0dde11623509/bundler/lib/bundler/plugin.rb#L228). I _was_ able to [trick bundler in to thinking a non-existing gem provided our rehash plugin](https://github.com/benmelz/rbenv/blob/c25bbc358c0838b737d87f494c15cef491a0ef15/rbenv.d/exec/gem-rehash/rubygems_plugin.rb#L17) by using an ugly hack (the `Struct`). That being said, you _could_ look at it as an ugly hack in a less important (thus more prone to changes) part of private API.
- And then additional reasons I discounted this when I first wrote the branch - there were some weird console output related to the plugin registration, and I was observing some odd files being created in local `.bundle` directories as well. Of course I'm not observing that behavior now, though, so there might have been something else going on.

If the former sounds okay to you though, I'd be happy to swap the implementations out. Just have to smooth out some more stuff with the "detect if the # of bins changed" logic.

mislav:
> * The bundler plugin API is heavily tied to the concept that gems provide plugins and plugins provide hooks (like `after-install-all`) - even if you try to programmatically add your own hooks, bundler will refuse to run them unless there's an associated plugin.

Ah, that does sound like some hoops to jump through and not at all simpler than monkeypatching. Thanks for giving it a shot!

rbenv.d/exec/gem-rehash/rubygems_plugin.rb
```diff
@@ -10,27 +10,27 @@
   end
 end
 
-if defined?(Bundler::Installer) && Bundler::Installer.respond_to?(:install) && !Bundler::Installer.respond_to?(:install_without_rbenv_rehash)
+if defined?(Bundler::Installer) && Bundler::Installer.methods.include?(:install) && !Bundler::Installer.methods.include?(:install_without_rbenv_rehash)
```
mislav, line 13: `Class.methods.include?` checks for the presence of a class method, but since we switched to monkeypatching `install` on the instance level, we should be checking method presence on the instance too

```suggestion
if defined?(Bundler::Installer) && Bundler::Installer.method_defined?(:install) && !Bundler::Installer.method_defined?(:install_without_rbenv_rehash)
```

mislav (changes requested):

mislav (approved):
Thank you!

PR #1544, opened 2023-12-23
Update link to maintained version of Bats

acook:
The old Bats was archived in 2021, the currently maintained fork uses a different repo which is not linked to from the old one.


hsbt:
👍  Thanks!

PR #1556, opened 2024-03-14
Add exec command to README

TWiStErRob:


mislav (commented):
Hi, thank you for the suggestion, but `rbenv exec` is sort of an "internal" command (it's used by rbenv shims) and normally never needs to be known or interacted with the user.

In the context of your example, the user would run `gem environment gemdir`, which would dispatch to the `gem` rbenv shim, and would never need to run `rbenv exec gem …`.

TWiStErRob:
What if the two commands display different values?

mislav:
If the two commands display different values, then it's likely that rbenv wasn't set up properly in one's shell; i.e. that running `gem` does **not** resolve to a shim in `~/.rbenv/shims/gem`. If that happens, check your PATH setting.

PR #1568, opened 2024-04-29
rbenv init: modify shell config files instead of printing instructions

mislav:
When running `rbenv init`, typically during rbenv setup, users expected their shell environment to be modified permanently. Instead, what the command would do is print the instructions to the user and expect them to edit their shell initialization files accordingly. This proved to be unintuitive.

Now, running `rbenv init <shells>...` will modify the shell initialization files of the following shells:

- bash: `~/.bash_profile` or `~/.bashrc` if the latter exists but the former does not
- zsh: `~/.zprofile` or `~/.zshrc` if the latter exists and mentions "rbenv"
- fish: `~/.config/fish/config.fish`

If no shells were specified on the command line, rbenv will try to detect the current shell.

It should be safe to run `rbenv init` multiple times, as the command will avoid modifying any shell startup file that already mentions "rbenv init".

### DIFF
--- a/libexec/rbenv-init
+++ b/libexec/rbenv-init
@@ -1,6 +1,19 @@
 #!/usr/bin/env bash
 # Summary: Configure the shell environment for rbenv
-# Usage: eval "$(rbenv init - [--no-rehash] [<shell>])"
+# Usage: rbenv init [--no-rehash] [<shells>...]
+#        rbenv init - [--no-rehash] [<shell>]
+#
+# Modifies shell initialization files to bootstrap rbenv functionality.
+# Typically, this will add a line that eval's the output of `rbenv init -`.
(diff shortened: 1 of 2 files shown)


mislav:
@baelter Caution should be taken with `.zshenv` because that script gets loaded in all zsh shells, no matter whether they are "login" or interactive. For instance, script files with `#!/bin/zsh` shebang would load `.zshenv` too when executed. Since rbenv is generally not needed in those instances, bootstrapping rbenv with `eval "$(rbenv init ...)"` would add unnecessary overhead.

Sadly, there is no one great place to add this bootstrapping code to that would work across all systems. The situation with bash is even worse. Users will need to adapt to the specifics of their shell and their system.

PR #1569, opened 2024-05-01
Use compsys completion system for zsh

Farid-NL:
Completions in zsh are done via [`compctl`](https://zsh.sourceforge.io/Doc/Release/Completion-Using-compctl.html).
This PR update the completion file to use the 'new' completion system: [`compsys`](https://zsh.sourceforge.io/Doc/Release/Completion-System.html).

The changeset doesn't include descriptions for commands, it just use the current logic implemented for `compctl` and use `compsys` instead.

### DIFF
--- a/README.md
+++ b/README.md
@@ -286,7 +286,7 @@ name | default | description
 
 1. Prepends `~/.rbenv/shims` directory to PATH. This is basically the only requirement for rbenv to function properly.
 
-2. Installs shell completion for rbenv commands.
+2. Installs bash shell completion for rbenv commands.
 
 3. Regenerates rbenv shims. If this step slows down your shell startup, you can invoke `rbenv init -` with the `--no-rehash` flag.
 
--- a/README.md
+++ b/README.md
@@ -103,6 +103,24 @@ This will get you going with the latest version of rbenv without needing a syste
 
 3. Restart your shell so that these changes take effect. (Opening a new terminal tab will usually do it.)
 
+#### Shell completions
+
+When _manually_ installing rbenv, it might be useful to note how completion scripts for various shells work. Completion scripts help with typing rbenv commands by expanding partially entered rbenv command names and option flags; typically this is invoked by pressing <kbd>Tab</kbd> key in an interactive shell.
+
+- The **bash** completion script for rbenv ships with the project and gets [loaded by the `rbenv init` mechanism](#how-rbenv-hooks-into-your-shell).
+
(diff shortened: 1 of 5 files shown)


mislav:
Thanks! I have pushed some changes:
- the script is now renamed to `_rbenv` to enable autoloading
- rbenv command completion now includes command descriptions

To try this out, please ensure that the project's `completions/` directory is in your zsh's `$fpath` before it initializes the completion system. When this ships, I imagine that it should be the responsibility of a package manager to put the `_rbenv` file into an appropriate location.

Any feedback welcome!

Farid-NL:
It works great! I've pushed a little change.

- `compedef` directive takes its arguments as commands to which it will complete, whence the change from `_rbenv` to `rbenv`.

It wasn't working for me until that.

Farid-NL:
> I imagine that it should be the responsibility of a package manager to put the _rbenv file into an appropriate location.

I would imagine the same as well, however, what about people using the [basic git checkout](https://github.com/rbenv/rbenv?tab=readme-ov-file#basic-git-checkout) installation method?

I suppose it could be indicated in the README that it is necessary to add project's `completions/` in the `$fpath`. Or tweak `rbenv-init`, to add the directory to `$fpath` only in zsh shells, as the commit above.

mislav:
> suppose it could be indicated in the README that it is necessary to add project's `completions/` in the `$fpath`.

That's a very good point. Users of “basic git checkout” installation method will have to tweak their zsh environment to get completions working again. This might be slightly disruptive, but I think proper autoloading is a better solution going forward. We can document this in the README and in future release notes.

> Or tweak `rbenv-init`, to add the directory to `$fpath` only in zsh shells, as the commit above.

That's a solid idea, but changing `$fpath` has no effect if the completion system has already been initialized. That means that zsh completions for rbenv would either work or not work depending on where `eval … rbenv init` line is in zsh startup files relative to completion initialization. Since this is a bit tricky, I would rather opt for users (or package managers) managing fpath themselves and not trying to interfere with that process from within rbenv. With that in mind, I might ship this changeset without your last commit. Would that be okay?

Farid-NL:
> Since this is a bit tricky, I would rather opt for users (or package managers) managing fpath themselves and not trying to interfere with that process from within rbenv. With that in mind, I might ship this changeset without your last commit. Would that be okay?

By all means. You're completely right, it might be tricky changing `$fpath` just like that.

PR #1604, opened 2025-01-07
Use readarray in bash v4+ to avoid rbenv init hanging

### LINKED ISSUE #1593: eval "$(rbenv init - zsh)" got stuck.
`eval "$(rbenv init - zsh)"`

image

mislav:
For just a handful of people, rbenv init would hang indefinitely. Turning on debugging output suggested that the `read` expression to split PATH into a bash array was hanging, but I could never reproduce this myself. Instead, this uses bash v4+ `readarray` if it's available, or falls back to bash v3 `read` with the default DELIM being a newline character. This will cause a regression if any PATH entries contain a literal newline character, but hopefully people are not relying on such paths.

Fixes #1593
Followup to #1450

### DIFF
--- a/libexec/rbenv-commands
+++ b/libexec/rbenv-commands
@@ -26,7 +26,12 @@ elif [ "$1" = "--no-sh" ]; then
   shift
 fi
 
-IFS=: read -d '' -r -a paths <<<"$PATH" || true
+if [ "$(type -t readarray)" = "builtin" ]; then
+  readarray -d : -t paths < <(printf "%s" "$PATH")
+else
+  # bash 3.x compatibility
+  IFS=: read -r -a paths <<<"$PATH" || true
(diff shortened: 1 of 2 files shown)


Tobias-pwnr:
I don think is a good fix for bash 4

```
 bash --version
GNU bash, version 4.2.46(2)-release (x86_64-redhat-linux-gnu)
Copyright (C) 2011 Free Software Foundation, Inc.
```

when I try  eval "$(rbenv init -)"

```...rbenv/libexec/rbenv-commands: line 30: readarray: -d: invalid option
readarray: usage: readarray [-n count] [-O origin] [-s count] [-t] [-u fd] [-C callback] [-c quantum] [array]
-bash: syntax error near unexpected token `)'
```

PR #1605, opened 2025-01-08
fix readarray -d invalid option

pnghai:
Only readarray @ Bash 4.4+ support -d option

https://github.com/rbenv/rbenv/pull/1604#issuecomment-2576749563

This fix allow Bash lower than 4.4 can still be worked with good old IFS read... instead


mislav:
Hi, thank you for the proposed fix. Your change looks good, but I'm working on an alternate solution; one that does not use readarray nor `read` at all, since both have proven to be unreliable across bash versions/systems.

PR #1613, opened 2025-03-20
Update README.md to use Ruby version 3.4.2 instead of 3.1.2

Kiran-Rao:
Update to show the latest ruby version.

Useful so people are less confused/accidentally set their environment to use an older version of ruby.


mislav:
Hi, thank you, but the current documentation serves as an example how to install any Ruby version, so it doesn't matter much that it's not the latest. People using Ruby for a project will already know which Ruby version they want, and I would prefer to avoid committing to always having the latest Ruby version in the README so that we don't have to update the docs too often.

PR #1619, opened 2025-04-23
Updating readme

gitKearney:
Making README a little more readable with a ToC. Making the installation section a little more clear


gitKearney:
Sorry for all the commits, but GitHub's preview wasn't working, or would make some links appear that they were not valid

mislav:
Hi, thanks for suggesting changes to the documentation, but I cannot accept a PR that completely rewrites the most important piece of documentation for this project, at least not without the change of this magnitude being warranted somehow.

The primary purpose of this PR seems to be to add a ToC so that users can jump to different sections of the document or even link to a specific sub-section, but I see to main problems with this:

1. This project did have a ToC but [we removed it](https://github.com/rbenv/rbenv/commit/b0fb351419c0688b58a05325fd251c0a08903aad) after GitHub shipped an automatic ToC (via a dropdown toggled by an icon) to every README document on the platform;

2. A bunch of headings got renamed in your version of the README, so whoever is currently linking to subsections of the README will find those links to the broken if I merged your changes. That regression would actually make the current ToS functionality worse.

Furthermore, a bunch of smaller style changes didn't resonate with me, particularly that subheadings are in UPPERCASE now and the reformatting of the command reference as a table. Even though a table is easier to scan, changing the format required deleting most of the prose describing how different rbenv commands work, and I'd rather keep the expanded documentation than having a terse table overview.

And finally, just a bit of advice since you're obviously well-intentioned with your contribution: we open source maintainers are often really welcoming of documentation improvements, especially because outside contributors might have better perspective on a piece of software than we have of our own projects. But, as a first time contributor to this project, you'd have a much better chance of getting a contribution accepted if it was a smaller change, as opposed to rewriting a whole document. Just something to think about for next time! Cheers

